PR #1, opened 2018-09-17
new pcairPartition function

mconomos:
I made a new R script with a new function called pcairPartition2. In the end, this should become the one true pcairPartition.

There are actually two functions currently: pcairPartition_gds and pcairPartition_matrix.  They are very similar; the only places they differ are where we access the gds (I think 5 lines in the code).  I'm hoping  you can add in the code to make these work with S4 methods. Ideally we could mix and match gds and matrix for the kinobj and divobj.


smgogarten (commented):
Combined two new functions into one by defining S3 methods for allowed object types. To-do:
- add utility functions to convert KING to GDS
- update pcair to have consistent argument names
- pcair probably needs a rewrite as well

R/pcair.R
```diff
@@ -1,299 +1,133 @@
...
+    out <- list(vectors = eigenvect,
+                values = eigenval,
+                rels = part$rels,
+                unrels = part$unrels,
                 kin.thresh = kin.thresh,
-                div.thresh = -abs(div.thresh),
-                nsamp = scan.include$n,
-                nsnps = nsnps,
-                MAF = MAF,
+                div.thresh = div.thresh,
```
mconomos, line 124: I'm not enforcing div.thresh to be negative in the function anymore, so this doesn't need to be wrapped in -abs()

R/pcair.R
```diff
@@ -1,299 +1,133 @@
...
-    # add scanIDs as rownames of EIG
-    rownames(EIG) <- scan.include$value
-    
-    # return results
-    out <- list(vectors = EIG, 
-                values = L, 
-                sum.values = sum.values, 
-                rels = rels, 
-                unrels = unrels,
+    # output object
```
mconomos, line 118: Would it make sense for us to include a vector of sample.id as part of the output. I know they are the rownames of the vectors, but is it useful to have it separate? I'm fine with it either way.

smgogarten: Yes, that's a good idea. It would be consistent with the fitNullModel output as well as all the SNPRelate functions.

mconomos:
I added a new file called kingToMatrix.R (and corresponding .Rd). This should end up just replacing `king2mat`, as it does the same thing but faster, and also can work with the --ibdseg output.
If you want to change the name back to `king2mat`, that's completely fine with me (or it can stay as `kingToMatrix`). I just didn't want to get rid of the old one until you had a chance to look at the new one

smgogarten:
kingToMatrix is enough different from king2mat that I think we should deprecate king2mat and keep the new name. We should have an example file with the --ibdseg output. Using the small test datasets that come with GENESIS or SNPRelate gives the error "No informative IBD segments", so maybe we should try a chromosome from 1000 genomes phase 3.

NAMESPACE
```diff
@@ -8,6 +8,7 @@ import(Biobase)
 import(gdsfmt)
 import(graph)
```
mconomos: i'm not sure that we're using stuff from the `graph` package anymore. I think that was being used in the old `pcair` function. I don't think it's being used anywhere else.

smgogarten: removed graph from DESCRIPTION and NAMESPACE

PR #3, opened 2018-10-11
Pcrelate

mconomos:
This version of pcrelate runs. It works with multiple blocks of variants and/or samples, but nothing is parallelized yet. I ran some small tests and can get the same output as the original pcrelate.

@smgogarten - please build in the variant iterator classes (should be two locations). We will need to discuss parallelization.


mconomos:
Adding parallelization over CPU cores with foreach and the doMC packages was shockingly easy. Running on my toy example, it not surprisingly made big improvements in performance.

Also, on a very small scale test, making more small blocks of variants run in parallel (e.g. 8 blocks of 50) vs. one larger block of variants (e.g. 1 block of 400) is quite a bit quicker. Most likely because it can read from the GDS in multiple places at once, as you suggested the other day.

I think the big things that need to be done are just adding in iterator classes and building in some more checks.

smgogarten (commented):
I added a unit test comparing the old and new versions, and the results don't match. Should I be using different arguments?

smgogarten:
Replaced snp block creation with use of iterator classes. Added support for GenotypeData as well as SeqVarData. Added unit tests to compare output to original version. To-do: decide on desired output format (maybe not data.table?), rewrite other functions (pcrelateMakeGRM, etc) to work with it. Also: should we keep both original and new versions in package for next release?

smgogarten:
Sorry about the merge commit, not sure why it did that since we weren't modifying the same files. Issues: 1) for pcrelateMakeGRM, using the makeSparseMatrix function results in a sparse matrix object even if the GRM is actually dense. 2) the messages printed by makeSparseMatrix don't always make sense when it's used in other functions. Without a threshold, it says everyone is related and creates a single cluster, even when it seems like they should be unrelated.

smgogarten:
I added some more unit tests to exercise all the possible arguments, and got an error with `scale="none"`:
```
>     myrel2 <- pcrelate(iterator, pcs = mypcs, scale="none", ibd.probs=FALSE, verbose=FALSE)
Error in diag(estList$kin) : 
  (list) object cannot be coerced to type 'double'
```

PR #4, opened 2018-10-16
Matrix

smgogarten:
reduce dependency on Matrix package, and only convert to Matrix objects in some cases.
streamline code for nullModelTestPrep and calcXtilde - no more special cases for glm.


R/nullModelAsMatrix.R
```diff
@@ -0,0 +1,27 @@
+.checkMatrixType <- function(covMatList) {
+    isMatrix <- sapply(covMatList, is, "Matrix")
+    if (any(isMatrix) & !all(isMatrix)) {
+        for (i in which(isMatrix)) {
```
mconomos: shouldn't this be for i in which they are not isMatrix (i'm having a hard time thinking how to write that in code at the moment)

mconomos:
Everything else looks good to me 👍

R/nullModelAsMatrix.R
smgogarten: Yes, it should.

PR #5, opened 2018-11-15
fixed sample ordering issue for running calcISAFBeta and added a check

mconomos:
Can you take a quick look at this; we need to update this for Caitlin to run pcrelate in the parallelized way.  I did two things:
1) Changed the way pcs are subset to sample.include in .createPCMatrix. The pcs used to be reordered by sample.include; now they are not.
2) Added sample order checks to .calcISAFBeta

### DIFF
--- a/R/pcrelate2.R
+++ b/R/pcrelate2.R
@@ -233,8 +233,7 @@ setMethod("pcrelate",
 ### function to match samples and create PC matrix
 .createPCMatrix <- function(pcs, sample.include){
     # subset and re-order pcs if needed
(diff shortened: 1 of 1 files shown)


smgogarten (changes requested):
Line 236 should stay as is, because users might not always have PCs in the same order as the GDS file. We should always call `.sampleInclude` before using `calcISAFBeta` and `pcrelateSampBlock` independently.
Checking that the sample ids match is good and important.

mconomos:
Line 236 has been reverted to what it was

PR #6, opened 2018-12-05
add option to read dosage from DS field instead of genotypes

smgogarten:
Questions:
1. Will Michigan's imputation server ever return >1 dosage value per variant?
2. Is "imputedDosage" (as opposed to refDosage and altDosage which read from genotypes) a good name for the function in SeqVarTools that GENESIS is calling?
2. is "imputed=FALSE" a good name for the argument that turns on reading from the "DS" field instead of counting alleles from the genotype field?

### DIFF
--- a/R/assocTestAggregate.R
+++ b/R/assocTestAggregate.R
@@ -7,7 +7,7 @@ setMethod("assocTestAggregate",
                    test=c("Burden", "SKAT", "SMMAT"),
                    burden.test=c("Score", "Wald"), rho=0,
                    pval.method=c("davies", "kuonen", "liu"),
-                   sparse=TRUE, verbose=TRUE) {
+                   sparse=TRUE, imputed=FALSE, verbose=TRUE) {
 
               # check argument values
               test <- match.arg(test)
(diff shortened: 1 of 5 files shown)


mconomos:
1. Does your question mean will the _value_ ever be greater than 1? Or can there ever be more than 1 value?  I'm not 100% sure, but I think it can return more than one value for a variant if there are multiple alternate alleles observed in the reference. This probably depends on some settings though, and I don't really know for sure.

2. I like the name "imputedDosage". The one thing this doesn't convey is if it's the imputed dosage or ref or alt alleles, but that's probably OK.

3. I think "imputed = FALSE" is good. Easily understandable and concise. The only other name I would have thought of would be something like "dosage = FALSE", but we use the term dosage to also mean counts of observed alleles.

sarahcn:
_corrected on # 1 about maximum DS value_
1. DS value itself should always be < _2_ [=P(0/1)+2*P(1/1)]. Regarding whether there can be more than 1 DS value per variant - I do not know (or cannot recall) how UMich has formatted multi-allelic vars (with > 1 ALT allele) in their reference panels. For IMPUTE2, multi-allelic vars were "decomposed" such that there would be one row of data for each ALT allele (effectively row1=REF, ALT1; row2=REF, ALT2, etc.). But I can't quickly determine that with some example UMich output, so let me check with them and cc you.

2. Looks good to me. Some variants will be imputed and some will be observed, but people still generally think of it as an "imputed" dataset.

3. Looks good for same reason as #2. Another option instead of "imputed=TRUE" (if that's what the flag is for the imputed option) might be "probs=TRUE" or "discrete=FALSE" to demarcate the fact that you're reading from a continuous dosage field rather than calculating dosage from discrete genotype calls. (If I understand correctly what this option is about).

Excuse the delayed response!

mconomos:
@sarahcn 
In (1) do you mean that DS should always be **_< 2_**? From the definition you wrote down, if DS = P(0/1) + 2*P(1/1), then when P(1/1) = 1, DS would be 2.  Right?

sarahcn:
@mconomos whoops - yes, that is what I meant - thanks for catching that. so strengthens the case that @smgogarten was asking about whether there can be more than one DS value per variant, not about what threshold that DS value may or may not exceed

PR #9, opened 2019-01-23
re-enable pcair to take only a user specified unrel.set as input

mconomos:
This will solve the problem that Leon sent us earlier. There are now a couple options to run pcair()

1) Specify kinobj and divobj (with or without unrel.set specified).  This is what the current version of the code requires.

2) Leave kinobj and divobj both as NULL. 
a) If unrel.set is specified and sample.include is NULL, it will treat these samples as the unrelated set, and any other samples in gdsobj as the related set.  
b) If unrel.set is specified and sample.include is also specified, it will treat any samples in unrel.set and sample.include as the unrelated set, and any samples not in unrel.set but in sample.include as the related.set.
c) If unrel.set is also NULL, it will do a standard PCA analysis on the samples in sample.include (or gdsobj if unspecified).

3) If one of kinobj or divobj is specified and the other is NULL, it will stop with an error message.

I ran some manual tests using your testthat code. We should probably build in a couple extra tests there.

### DIFF
--- a/R/pcair.R
+++ b/R/pcair.R
@@ -4,8 +4,8 @@ setOldClass("gds.class")
 setMethod("pcair",
           "gds.class",
           function(gdsobj, 
-                   kinobj,
-                   divobj,
+                   kinobj = NULL,
+                   divobj = NULL,
                    kin.thresh = 2^(-11/2),
                    div.thresh = -2^(-11/2),
                    unrel.set = NULL,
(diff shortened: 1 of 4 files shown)


smgogarten (changes requested):
Update pcair.Rd with description of new options.

I think this is good, as it means that users can now just call pcair(gdsobj) and get a result.

smgogarten (approved):

PR #14, opened 2019-03-06
filter monomorphic variants

smgogarten:
Variants with all heterozygotes (clearly artifacts) were causing
spurious hits with p-value 0. Filter those variants, as well as
filter monomorphic variants (MAF 0) before running tests rather
than setting their values to NA after.


mconomos (approved):
All looks good to me.  Thanks for taking care of this.

PR #15, opened 2019-03-08
Update to inverse normalization to match Tamar's method

mconomos:
When norm.option == "by.group", everything should still work the same.

When norm.option == "all", I've allowed rescaling to happen, and it's done by the SD of the residuals across all samples.

This norm.option == "all" setup is what Tamar's method is for pooled analysis; even when there are multiple groups. I propose that we should do this as our default. We still need to update the analysis pipeline code to allow this.  Currently, the pipeline determines which norm.option to use by whether or not there is a group variable; we don't want to do this now.  We probably have to add a new optional argument, something like "norm.option" to allow users to do the inv-normalization and rescaling either by.group as we have been (we probably shouldn't remove that option), or using all samples. 

It seems like the conditional statements at the end of the null_model script can simplify down to: if doing inverse normalization, use the specified norm.option and rescale variables, where the default are "all" and "residSD" by default.

### DIFF
--- a/R/updateNullModOutcome.R
+++ b/R/updateNullModOutcome.R
@@ -3,7 +3,13 @@
 
 ## the names of items in the list group.idx have to match the names of the corresponding variance components!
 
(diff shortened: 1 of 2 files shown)


smgogarten (approved):
looks good

PR #18, opened 2019-05-04
updated kingToMatrix to account for new KING --related output

mconomos:
The function used to infer if it was --ibdseg or --kinship output from the column names present. However, with the --related option in KING now, that output contains both. The simplest solution is just to make the user specify which they want (they should know anyways), so I added a new 'estimator' argument. 

I'm not 100% sure I did the match.arg() thing correctly, so if you could at least check that over, that would be good. Thanks!


smgogarten (approved):
Added a check that the requested estimator is actually present in the file, otherwise it will issue a warning and return a diagonal matrix. Updated unit tests and examples to use `estimator` argument.

mconomos:
Thanks for catching that case!

PR #21, opened 2019-05-31
fastSKAT

smgogarten:
Add fastSKAT option to assocTestAggregate. This involves a rewrite of the variantSetTests code where SKAT is no longer a special case of SKAT-O, but shares code with fastSKAT.

This update includes code from the [bigQF](https://github.com/tslumley/bigQF) package.


mconomos (commented):
everything looks good. I merged the pull request

PR #25, opened 2019-07-16
MAC

smgogarten:
Include minor allele count in GENESIS output, calculated correctly for the sex chromosomes.

### DIFF
--- a/R/assocTestAggregate.R
+++ b/R/assocTestAggregate.R
@@ -70,21 +70,21 @@ setMethod("assocTestAggregate",
                                       male.diploid=male.diploid, genome.build=genome.build)
                   
                   # filter monomorphic variants
-                  keep <- .filterMonomorphic(geno, count=n.obs, freq=freq, imputed=imputed)
+                  keep <- .filterMonomorphic(geno, count=n.obs, freq=freq$freq, imputed=imputed)
 
                   # exclude variants with freq > max
-                  keep <-  keep & freq <= AF.max
+                  keep <-  keep & freq$freq <= AF.max
                   if (!all(keep)) {
                       var.info <- var.info[keep,,drop=FALSE]
                       geno <- geno[,keep,drop=FALSE]
                       n.obs <- n.obs[keep]
-                      freq <- freq[keep]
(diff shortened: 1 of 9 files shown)


R/assocTestAggregate.R
```diff
@@ -33,6 +35,9 @@ setMethod("assocTestAggregate",
               # do we need to match on alleles?
               match.alleles <- any(c("ref", "alt") %in% names(mcols(currentRanges(gdsobj))))
 
+              # check ploidy
+              if (SeqVarTools:::.ploidy(gdsobj) == 1) male.diploid <- FALSE
```
mconomos, line 39: @smgogarten should there be the opposite check here? i.e. if ploidy == 2, male.diploid <- TRUE ?

mconomos:
I think this all looks good. I made one minor comment in the code on the check.

mconomos (commented):
This all looks good to me. I made one minor comment on one of the checks. The same question applies to other places in the code where that check is done.

R/assocTestAggregate.R
smgogarten: Not necessarily. On the X chromosome, often ploidy == 2 because of the PAR, even though the non-PAR variants can still be coded as haploid in some cases. Have not yet solved the problem of how to return dosages as non-missing for 1000 genomes variants (coded as 0/. and 1/.), but this should work if the VCF is coded as 0/0 and 1/0.

mconomos: Ah. Good point. I think this looks good then. 

Do we need some sort of warning or disclaimer about this 0/. issue? What do the results end up looking like? From looking at the code, I'm guessing M.count is NA, which makes both freq and MAC also be NA. If that's true, at least it's not giving _wrong_ results.

smgogarten: In the case of the 1000 genomes X chrom file, it returns NA only for the males, so all (non-PAR) X chrom analyses will end up treating the males as missing. That is wrong, and I think we need to fix it rather than giving a disclaimer. But the fix needs to be in SeqArray and SeqVarTools rather than GENESIS.

PR #27, opened 2019-07-25
update to makeSparseMatrix to work on large matrices

mconomos:
In trying to make the sparse matrix for freeze 5, I discovered that makeSparseMatrix didn't work on matrices with > 2^31 elements.... 🤦‍♂ 

I came up with a different way to do this. As a result, matrix/Matrix objects are now handled differently than data.frame/data.table objects. The function .makeSparseMatrix_df is the same as what .makeSparseMatrix used to be. The new function .makeSparseMatrix_matrix is a modified version of that function to work matrix (or Matrix) objects. These are all still called as methods for makeSparseMatrix. If you want to do this a different way, that's fine with me. I tested the new function on the fr5 KM, and it seems to work. 

I moved meltMatrix to the pcrelate file simply because it's only used there now.

### DIFF
--- a/R/makeSparseMatrix.R
+++ b/R/makeSparseMatrix.R
@@ -36,6 +36,18 @@ setMethod("makeSparseMatrix",
         stop("thresh must be specified when x is a matrix object")
     }
 
+    # check for names
+    if(is.null(colnames(x))) {
+        if(is.null(rownames(x))) {
+            stop("dimnames must be provided for matrix x to track sample order in the output matrix.")
+        } else {
+            colnames(x) <- rownames(x)
+        }
+    }
+    if(is.null(rownames(x))) {
+        rownames(x) <- colnames(x)
+    }
+    
     # check that matrix is square
     if(!all(rownames(x) == colnames(x))){
         stop("x must be square when it is a matrix object")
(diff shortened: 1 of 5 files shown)


smgogarten:
There is no `diag.value` option in `makeSparseMatrix_matrix`, but kingToMatrix still uses it: https://github.com/UW-GAC/GENESIS/blob/makesparse/R/makeSparseMatrix.R#L279

mconomos:
@smgogarten thanks for catching that; I was thinking that KING would only be coming from the actual software, and hence not a matrix.

It was easy enough to put the `diag.value` argument back into the `makeSparseMatrix_matrix` function. See the new commit.

smgogarten:
The block starting at L61 fails if the input matrix has no dimnames. Maybe the simplest thing is to assign names if none exist? That would involve copying the matrix object though, which could get expensive for very large matrices.
https://github.com/UW-GAC/GENESIS/blob/makesparse/R/makeSparseMatrix.R#L61

mconomos:
Hmmm... I see why that would fail if it has no names. However, I don't see how to make the code there work without names without modifying a bunch of stuff downstream as well. While it may not be the nicest user experience, what do you think about adding a check at the beginning to see _if_ the matrix has dimnames, and stopping if it does not; i.e. requiring that the user assign names. The function wouldn't be all that useful without names in the first place, seeing as it re-orders samples to be block diagonal.

smgogarten (commented):
The default value of "thresh" shouldn't be NULL for the matrix methods if that's not an allowed option. Should we have a default thresh value? Maybe `2^(-11/2)` for consistency with pcairPartition?

mconomos:
Good point; `2^(-11/2)` works for me.

PR #31, opened 2019-08-27
count non missing

mconomos:
is.na can not be applied to Matrix objects with > 2^31 elements. This breaks our code for counting the number of observed genotypes in a few places. 

Added a .countNonMissing function to utils.
Updated assocTestAggregate and .alleleFreq to use this function.


smgogarten (approved):
Looks good. I made equivalent changes in assocTestSingle.

PR #32, opened 2019-09-06
Carriers only tests

jiwonslee:
Added carriers only tests to assocTestSingle.R and assocTestAggregate.R, included split methods, and utils for split methods.


smgogarten:
Please remove the ".o" files from github; they are a result of compiling the C code and should be ignored. (One could make a `.gitignore` file for them, but I haven't done it yet.)

smgogarten:
I think all functions (especially exported functions) should have consistent naming conventions, which means camelCase rather than snake_case.

smgogarten (changes requested):
Please add unit tests for the new test methods (BinomiRare and CMP) and the split functions.

Also, I think the man pages need more details on what the BinomiRare and CMP tests are doing.

smgogarten:
The argument "max.alt.freq" to assocTestSingle is essentially the same as the "AF.max" argument to assocTestAggregate. I think they should be consistent, and though I actually like your name better (as it's more clear), I think it might be too disruptive to existing code to change the name now.

R/assocTestAggregate.R
```diff
@@ -1,252 +1,264 @@
...
+    if (length(test) > 1) test <- NULL
+    match.arg(test, choices=c("Burden", "SKAT", "fastSKAT", "SMMAT", "fastSMMAT", "SKATO"))
+}
+
+setMethod("assocTestAggregate",
+          "SeqVarIterator",
+          function(gdsobj, null.model, AF.max=1,
+                   weight.beta=c(1,1), weight.user=NULL,
+                   test=c("Burden", "SKAT", "fastSKAT", "SMMAT", "SKATO"),
+                   burden.test=c("Score", "Wald", "BinomiRare","CMP"),
```
mconomos, line 13: I had planned to remove the `burden.test` argument. The plan is to always do the score test, and write a function that approximates the effect size, its standard error, and percent variance explained. I'm undecided at this point if those 3 values will always be included in the main output, or if it will be a post-processing function that users can apply to a subset of results. 

I'm thinking that "BinomiRare" and "CMP" could just be options to the `test` argument. Do you see any reason that that would not work?

R/assocTestAggregate.R
```diff
@@ -1,252 +1,264 @@
...
+                      freq <- freq[keep,,drop=FALSE]
+                  }
+
+                  # weights
+                  if (is.null(weight.user)) {
+                      # Beta weights
+                    if (burden.test %in% c("CMP", "BinomiRare")){
+                      weight <- seq(from=1, to=1, length.out=length(freq$freq))
+                    } else {
+                      weight <- .weightFromFreq(freq$freq, weight.beta)
```
mconomos: We probably need to make it explicit that CMP and BinomiRare ignore the `weight.beta` argument.

R/assocTestAggregate.R
```diff
@@ -1,252 +1,264 @@
...
+                      n.obs <- n.obs[keep]
+                      freq <- freq[keep,,drop=FALSE]
+                  }
+
+                  # weights
+                  if (is.null(weight.user)) {
+                      # Beta weights
+                    if (burden.test %in% c("CMP", "BinomiRare")){
+                      weight <- seq(from=1, to=1, length.out=length(freq$freq))
+                    } else {
```
mconomos, line 89: We probably need to make it explicit that CMP and BinomiRare ignore the `weight.beta` argument

R/assocTestSingle.R
```diff
@@ -1,132 +1,151 @@
...
-          })
+setGeneric("assocTestSingle", function(gdsobj, ...) standardGeneric("assocTestSingle"))
+
+## do we want the GxE.return.cov option?
+## do we want to make imputing to the mean optional?
+setMethod("assocTestSingle",
+          "SeqVarIterator",
+          function(gdsobj, null.model,  test=c("Score", "Wald",  "SAIGE", "BinomiRare", "CMP"), 
+                   GxE=NULL, sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), 
+                   AF.max=NULL, calc.score=FALSE, verbose=TRUE) {
```
mconomos, line 9: It seems odd to me to have an AF.max argument on single variant tests. What is the motivation behind this?

R/assocTestAggregate.R
```diff
@@ -1,252 +1,264 @@
...
+          })
+
+
+
+setMethod("assocTestAggregate",
+          "GenotypeIterator",
+          function(gdsobj, null.model, AF.max=1,
+                   weight.beta=c(1,1), weight.user=NULL,
+                   test=c("Burden", "SKAT", "fastSKAT", "SMMAT", "SKATO"),
+                   burden.test=c("Score", "Wald", "BinomiRare","CMP"),
```
mconomos, line 155: Same comment as above

R/assocTestAggregate.R
```diff
@@ -1,252 +1,264 @@
...
+
+                  # weights
+                  if (is.null(weight.user)) {
+                      # Beta weights
+                      weight <- .weightFromFreq(freq$freq, weight.beta)
+                      if (burden.test %in% c('BinomiRare', 'CMP')){
+                        weight <- seq(from=1, to=1, length.out=length(weight))
+                      }
+
+
```
mconomos, line 209: Same comment as above

R/assocTestSingle.R
```diff
@@ -1,132 +1,151 @@
...
+          })
+
+
+
+setMethod("assocTestSingle",
+          "GenotypeIterator",
+          
+          
+          function(gdsobj, null.model,  test=c("Score", "Wald",  "SAIGE", "BinomiRare", "CMP"), 
+                   GxE=NULL,male.diploid=TRUE, AF.max=NULL, calc.score=FALSE, verbose=TRUE) {
```
mconomos, line 90: Same comment about AF.max as above

R/testGeno.R
```diff
@@ -1,215 +1,325 @@
...
+    if (test == "Wald" & !is.null(E)){
+        res <- .testGenoSingleVarWaldGxE(nullmod, G, E, GxE.return.cov.mat=GxE.return.cov)
+    }
+    
+    if (test == "Score"){
+        Gtilde <- calcGtilde(nullmod, G)
+        res <- .testGenoSingleVarScore(Gtilde, G, nullmod$resid)
+    }
+    
+
```
mconomos, line 41: SAIGE should only be used for binary outcomes, right? we need to add a check for that if it's correct. I also don't think this option should be called "SAIGE". SAIGE is the UM's software that does many things to make fast and accurate tests. The particular thing implemented here is the saddlepoint approximation (SPA) to calibrate score tests with unbalanced case control ratios. 

I actually sort of think SAIGE should not be an option to "test", but something like "SPA" should be a logical option when using test = "Score".

R/variantSetTests.R
```diff
@@ -1,458 +1,480 @@
...
+
+## function that gets an n \times p matrix of p genotypes of n individuals, and a null model, and tests the genotypes associations with the outcomes. 
+## Genetic data are always assumed complete. 
+## Types of tests: 
+## Variant set: SKAT, burden, SKAT-O. Multiple types of p-values. Default: Davies with Kuonen if does not converge. 
+
+
+testVariantSet <- function( nullmod, G, weights, 
+                            test = c("Burden", "SKAT", "fastSKAT", "SMMAT", "fastSMMAT", "SKATO"),
+                            burden.test = c("Score", "Wald", "BinomiRare", "CMP"),
```
mconomos, line 10: Again, I think BinomiRare and CMP should be options to the `test` input

PR #33, opened 2019-09-27
Saige

mconomos:


R/assocTestSingle.R
```diff
@@ -4,7 +4,7 @@ setGeneric("assocTestSingle", function(gdsobj, ...) standardGeneric("assocTestSi
 ## do we want to make imputing to the mean optional?
 setMethod("assocTestSingle",
           "SeqVarIterator",
-          function(gdsobj, null.model, test=c("Score", "Wald"), GxE=NULL, sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), verbose=TRUE) {
+          function(gdsobj, null.model, test=c("Score", "Wald"), SPA = FALSE, GxE=NULL, sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), verbose=TRUE) {
```
mconomos: @smgogarten what are your thoughts on having `SPA` as a logical argument here? Alternatively, we could just have it as another option for `test`. The reason I did it this way as that using the `SPA` option still actually does a score test, it just does some p-value adjustment afterwards.

R/spa.R
```diff
@@ -0,0 +1,83 @@
...
+	if (!requireNamespace("SPAtest")) stop("package 'SPAtest' must be installed to calculate SPA p-values")
+	expit <- function(x){exp(x)/(1+exp(x))}
+
+	# add columns to score.results
+	score.result$SPA.pval <- NA_real_
+	score.result$SPA.converged <- NA_real_
+
+	# index for variants with Score.pval < pval.thresh
+	# only bother running SPA on these variants
+	idx <- which(score.result$Score.pval <= pval.thresh)
```
mconomos, line 9: @smgogarten I took this idea from the original saige code. Basically, it only does the SPA adjustment for variants with score pval < pval.thresh. I think this makes sense from a speed perspective. However, it does lead to the `SPA.pval` output in the results having `NA` values for the variants not tested. That may not be ideal for post processing qq plots and such. 

We could just copy `Score.pval` over into `SPA.pval` for those variants not run, but perhaps that's misleading. What do you think?

R/assocTestSingle.R
smgogarten: Are we planning on dropping the "test" argument entirely and only doing score tests? If so, then I think the "SPA" logical makes sense here.

R/spa.R
smgogarten: I would copy Score.pval into SPA.pval, because everyone will want to look at all the p-values together. We should just make clear in the documentation that it is happening.

R/assocTestSingle.R
mconomos: In the short term, I think yes. But a "test" argument might be the best way to incorporate the CMP and BinomiRare tests. 

One thing with SPA is that it first runs the exact same score test code, then does post processing (as opposed to binomirare and cmp which have their own testing code). Additionally, the way it's set up now, using SPA will return both the original `Score.pval` and a new `SPA.pval`. I didn't see a reason to throw out the `Score.pval`, but I suppose we could. That would also address my question below.

smgogarten: Is it possible to combine SPA and BinomiRare or CMP? If the SPA=T/F argument only makes sense for one of the `test` options, then it might make more sense to have it be its own `test` option. But I could go either way.

R/spa.R
mconomos: Should we do that or overwrite the `Score.pval` with the `SPA.pval`? This ties into my question above as well. If we overwrote, then the `SPA.converged` flag would actually be an indicator of if SPA ran and converged (`TRUE`), ran but didn't converge ('FALSE'), or wasn't run ('NA'). We could only overwrite if converged == `TRUE`

smgogarten: Would anyone ever want both p-values to compare them? If no, then overwriting makes sense. We could also check what SMMAT returns, for consistency - I think it gives multiple p-values. Can you do SPA for a burden test, and if so do we need to add this argument to assocTestAggregate as well?

R/assocTestSingle.R
mconomos: From looking at the BR and CMP code, it appears they have the option to just return the score pvalue if it is > 0.05 (the `calc.score` argument seems to determine this; if `TRUE`, get score pvalue, keep if > 0.05, compute BR pval if < 0.05; if `FALSE`, just compute the BR pval for every variant). This is actually similar to the way SPA is currently implemented with `pval.thresh`. **It seems to me that BR and CMP are alternatives to SPA.** 

Given this parallelism, perhaps we should treat them all similarly. We could have a `test` argument with values "Score", "Score.SPA", "BinomiRare", and "CMP".  

We could also then have a common `pval.thresh` (or better name) parameter that applies to "Score.SPA", "BinomiRare", and "CMP", giving the threshold for which you re-compute the p-value using the more sophisticated method. Setting the value to 1 would apply the method to all variants. (Or we could just set this to 0.05 and not make it user defined.)

R/spa.R
mconomos: Would anyone want both? probably not in production; possibly for testing. I think I'm leaning toward overwriting now. The BR and CMP code overwrite the score p-values when that option is selected. 

SMMAT returns multiple p-values: the burden p-value, the "SKAT" like p-values, and then the combined SMMAT p-value. This makes more sense here because they are all p-values for different tests. 

I guess I don't see any reason why you couldn't apply SPA to burden; the genotype vector would just be replaced by the burden vector. They seem to have done a similar thing with BR and CMP. I guess I should add this here as well.

R/assocTestSingle.R
smgogarten: I named this argument `recalc.pval.thresh`. Do you like that or can you think of a better name?

R/spa.R
smgogarten: Tamar says that SPA doesn't apply to the burden test: "The p-value computation under SPA uses the allele frequency. This is irrelevant in burden test. We originally tried this as well, but realized it’s inappropriate and while sometimes it looks okay sometimes it give bad results (because it’s not suppose to work)."

R/assocTestAggregate.R
```diff
@@ -148,15 +148,15 @@ setMethod("assocTestAggregate",
           function(gdsobj, null.model, AF.max=1,
                    weight.beta=c(1,1), weight.user=NULL,
                    test=c("Burden", "SKAT", "fastSKAT", "SMMAT", "SKATO"),
-                   burden.test=c("Score", "Wald"),
+                   # burden.test=c("Score"),
```
smgogarten, line 151: @mconomos what do you think about removing this argument since it only has one option now?

R/assocTestSingle.R
mconomos: I think I like this name. I've been having a hard time coming up with one that I like. 

Once we include the BinomiRare and CMP tests, what would the description be? Something like: "When test != 'Score', variants with a Score p-value below this threshold will have their p-value recalculated using the specified test"

R/assocTestAggregate.R
mconomos: I was originally thinking we would leave it because BinomiRare and CMP would be options once we incorporate them. However, looking at that code again, I think we could just make them `test` options. Whatever you think would be better.

R/assocTestSingle.R
```diff
@@ -4,7 +4,8 @@ setGeneric("assocTestSingle", function(gdsobj, ...) standardGeneric("assocTestSi
 ## do we want to make imputing to the mean optional?
 setMethod("assocTestSingle",
           "SeqVarIterator",
-          function(gdsobj, null.model, test=c("Score", "Wald"), GxE=NULL, sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), verbose=TRUE) {
+          function(gdsobj, null.model, test=c("Score", "Score.SPA"), recalc.pval.thresh=1, GxE=NULL,
```
mconomos: Should we make the default for `recalc.pval.thresh` 0.05?  Adrienne's results showed that for variants with p > 0.05, there was pretty much no difference at all.

R/variantSetTests.R
```diff
@@ -7,20 +7,20 @@
...
     test <- match.arg(test)
-    burden.test <- match.arg(burden.test)
+    # burden.test <- match.arg(burden.test)
     # pval.method <- match.arg(pval.method)
 
     G <- .genoAsMatrix(nullmod, G)
 
     if (test == "Burden") {
-        out <- .testVariantSetBurden(nullmod, G, weights, burden.test)
+        out <- .testVariantSetBurden(nullmod, G, weights, burden.test = "Score")
```
mconomos, line 23: should the burden.test option just go away here too?

R/variantSetTests.R
```diff
@@ -55,14 +55,15 @@ testVariantSet <- function( nullmod, G, weights,
...
     Gtilde <- calcGtilde(nullmod, burden)
+    RSS0 <- as.numeric(crossprod(nullmod$Ytilde))
     
     if (burden.test == "Score") {
-        out <- .testGenoSingleVarScore(Gtilde, G = burden, resid = nullmod$resid) 
-    }
-    if (burden.test == "Wald"){
-        out <- .testGenoSingleVarWald(Gtilde, Ytilde = nullmod$Ytilde,
-                                      n = length(nullmod$Ytilde), k = ncol(nullmod$model.matrix))
+        out <- .testGenoSingleVarScore(Gtilde, G = burden, resid = nullmod$resid, RSS0 = RSS0)
```
mconomos: and the if statement here could just go away, right?

R/testGeno.R
```diff
@@ -2,39 +2,42 @@
...
-    }
-        
-    if (test == "Wald" & !is.null(E)){
-        res <- .testGenoSingleVarWaldGxE(nullmod, G, E, GxE.return.cov.mat=GxE.return.cov)
+
+    # run the test
+    if(test == "Score"){
+        Gtilde <- calcGtilde(nullmod, G)
+        RSS0 <- as.numeric(crossprod(nullmod$Ytilde))
+        res <- .testGenoSingleVarScore(Gtilde, G, nullmod$resid, RSS0)
```
mconomos: I had asked this before, but it might have gotten lost in the shuffle of merging branches (or I missed your response). Is there a reason not to put the Gtilde and RSS0 calculations in the `.testGenoSingleVarScore` function? The only reason I can think it would be is if passing in the whole nullmod is a bad idea.

R/assocTestAggregate.R
mconomos: The more I think about it, the more I lean toward making those `test` options.

man/assocTestSingle.Rd
```diff
@@ -76,7 +84,12 @@
     %When \code{GxE} is not \code{NULL}, if \code{ivar.return.betaCov} is \code{TRUE}, then the output is a list with two elements.  The first, "results", is the data.frame described above.  The second, "betaCov", is a list with length equal to the number of rows of "results", where each element of the list is the covariance matrix of the effect size estimates (betas) for the genotype and genotype interaction terms.
 }
 
-\author{Matthew P. Conomos, Stephanie M. Gogarten, Tamar Sofer, Ken Rice, Chaoyu Yu}
+\references{
+    Dey, R., Schmidt, E. M., Abecasis, G. R., & Lee, S. (2017). A fast and accurate algorithm to test for binary phenotypes and its application to PheWAS. The American Journal of Human Genetics, 101(1), 37-49.
+    Zhou, W., Nielsen, J. B., Fritsche, L. G., Dey, R., Gabrielsen, M. E., Wolford, B. N., ... & Bastarache, L. A. (2018). Efficiently controlling for case-control imbalance and sample relatedness in large-scale genetic association studies. Nature genetics, 50(9), 1335.
+}
+
+\author{Matthew P. Conomos, Stephanie M. Gogarten, Tamar Sofer, Ken Rice, Chaoyu Yu, Wei Zhou}
```
mconomos: Should Wei Zhou also be removed here?

R/nullModelTestPrep.R
```diff
@@ -27,7 +27,10 @@ nullModelTestPrep <- function(nullmod){
         # resid <- nullmod$resid.marginal*C^2
     }
 
-    return(list(Ytilde = Ytilde, resid = resid, CX = CX, CXCXI = CXCXI))
+    # compute residual sum of squares under the null model
+    RSS0 <- as.numeric(crossprod(Ytilde))
```
mconomos, line 31: added RSS0 calculation to null model here

R/testGeno.R
```diff
@@ -2,39 +2,46 @@
...
-    if (test == "Wald" & !is.null(E)){
-        res <- .testGenoSingleVarWaldGxE(nullmod, G, E, GxE.return.cov.mat=GxE.return.cov)
+
+    # run the test
+    if(test == "Score"){
+        Gtilde <- calcGtilde(nullmod, G)
+        if(is.null(nullmod$RSS0)){
+            nullmod$RSS0 <- as.numeric(crossprod(nullmod$Ytilde))
+        }
+        res <- .testGenoSingleVarScore(Gtilde, G, nullmod$resid, nullmod$RSS0)
```
mconomos, line 34: took your suggestion for checking if RSS0 is in the null model already or not

R/assocTestSingle.R
smgogarten: The description is currently `If test is not "Score", only recalculate p-values below this threshold.` which seems general. But we can update it later to specifically mention the mulltiple test possibilities.

R/variantSetTests.R
smgogarten: We will still have to pass a `burden.test` argument to select BinomiRare and CMP in the future, since those (unlike SPA) can be used in the aggregate tests.

R/variantSetTests.R
smgogarten: When we add the carrier tests, there will still be if statements here:
`if (burden.test == "BinomiRare") out <- .testGenoSingleVarBR()`

PR #34, opened 2019-10-18
Remove Wald from single variant and burden tests

mconomos:


R/testGeno.R
```diff
@@ -2,65 +2,42 @@
...
+ 
     if(test == "Score.SPA" & nullmod$family$family != "binomial"){
         test <- "Score"
         message("Saddlepoint approximation (SPA) can only be used for binomial family; using Score test instead.")
     }
 
-
     # run the test
     if(test == "Score"){
         Gtilde <- calcGtilde(nullmod, G)
```
mconomos, line 28: @smgogarten Is there a reason we kept the `calcGtilde` calculation here instead of in `.testGenoSingleVarScore`? This and the new `RSS0` calculation could each go inside of that function, which would just require `G` and `nullmod` as input then. It would save us from repeating the 2 lines for both Score and Score.SPA.

Also, the new `RSS0` could actually just be computed and saved in the null model, but I'm not sure if that's a bad idea, because it would make the new code incompatible with old null models. What do you think?

R/testGeno.R
```diff
@@ -102,31 +79,34 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
...
+#     RSS <- as.numeric((sY2 - GPY * beta)/(n - k - 1))
+#     Vbeta <- RSS/GPG
+#     Stat <- beta/sqrt(Vbeta)
+#     res <- data.frame(Est = beta, Est.SE = sqrt(Vbeta), Wald.Stat = Stat, 
+#                       Wald.pval = pchisq(Stat^2, df = 1, lower.tail = FALSE))
+#     return(res)
+# }
 
 
 .testGenoSingleVarWaldGxE <- function(nullmod, G, E, GxE.return.cov.mat = FALSE){
```
mconomos, line 112: @smgogarten GxE is still "Wald"...it's a bit trickier to change this. I think I probably could, but I'm not sure if we should.

man/assocTestAggregate.Rd
```diff
@@ -83,12 +84,15 @@
     \item{Score.SE}{The estimated standard error of the Score}
     \item{Score.Stat}{The score Z test statistic}
     \item{Score.pval}{The score p-value}
-    If \code{burden.test} is \code{"Wald"}:
-    \item{Est}{The effect size estimate for a one unit increase in the burden value}
-    \item{Est.SE}{The estimated standard error of the effect size estimate}    
-    \item{Wald.Stat}{The Wald Z test statistic}
-    \item{Wald.pval}{The Wald p-value}
-    %% If \code{burden.test} is "Firth":
+    \item{Est.approx}{An approximation of the effect size estimate for each additional unit of burden}
```
mconomos, line 87: @smgogarten what do you think of these names? I actually don't really like "Est" anymore, but that would be consistent with what we used to return from the Wald test. I think I'd prefer something like "EffSize" or even "beta", but it might be best to keep it as similar as possible.

mconomos: We could also just make the names Est, Est.SE, and PVE. Est and Est.SE are what Wald used to return. The calculation of Est is actually identical to what was returned by Wald; Est.SE is very similar but not identical. PVE is new either way.

R/assocTestAggregate.R
```diff
@@ -10,7 +10,7 @@ setMethod("assocTestAggregate",
           function(gdsobj, null.model, AF.max=1,
                    weight.beta=c(1,1), weight.user=NULL,
                    test=c("Burden", "SKAT", "fastSKAT", "SMMAT", "SKATO"),
-                   burden.test=c("Score", "Wald"),
```
smgogarten, line 13: Are we leaving the burden.test argument here in anticipation of adding BinomiRare and CMP? Or should those go in the "test" argument instead? https://github.com/UW-GAC/GENESIS/pull/32#discussion_r326760137

R/testGeno.R
smgogarten: I think `calcGtilde` is separate because it takes a null model object, whereas the Score/Wald test functions just take matrices and vectors. Theoretically it would testing easier, but I can't actually find anywhere where we test `.testGenoSingleVarScore`directly, so it should not actually matter.

We could set the stage for moving RSS0 in the future by adding it to the null model now, but not expecting it to always be there (`if (is.null(nullmod$RSS0)) <calculate it>`)

R/testGeno.R
smgogarten: I think it's ok to leave it.

man/assocTestAggregate.Rd
smgogarten: I would keep the names the same as they used to be, it's minimally disruptive.

smgogarten (approved):

PR #35, opened 2019-10-22
added check for being a mixed model to varCompCI

mconomos:
I originally thought this was going to be more complicated...


smgogarten (approved):
Well, that's simple.

PR #39, opened 2020-05-15
Force design matrices to be non-sparse

amstilp:
The Matrix package by default creates a sparse matrix when more than 50% of the values are zero. For sparse matrices with only slightly more than 50% zero values, matrix multiplication is much less efficient. It is possible that a design matrix will have more than 50% zero values (e.g., if a large number of conditional variants are included), and thus be subject to reduced efficiency when fitting the null model, but it is unlikely that they will have enough zero values to make sparse matrix multiplication efficient when fitting the null model. This pull request forces all design matrices created with the Matrix package to be non-sparse and improves efficiency when fitting null models with a lot of conditional variants.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,1 +1,3 @@
+^renv$
+^renv\.lock$
 ^\.travis\.yml$
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,12 @@
 *.html
 *.so
 *.o
+
+# renv for development
+.Rprofile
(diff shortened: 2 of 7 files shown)


amstilp:
It looks like there are a lot of changes, but most are whitespace changes due to different text editor defaults.

mconomos (approved):
All looks good. Thanks for finding and fixing this.

smgogarten (approved):
Looks good, thanks!

PR #44, opened 2020-08-04
Bugfix/single sample blocks

sstadick:
This PR fixes an issue that pops up when one of your blocks only has a single sample in it. When `drop=FALSE` is not used in matrix indexing, and the vector you are using to index with has only one element, the row/col names will be dropped which eventually leads to a dataframe with non data due to how the `Reduce / merge` function works out. 

All tests pass for this, and I have confirmed this fixes my original issues.


sstadick:
Hello @smgogarten, just wanted to check in and see if you've had a chance to look at this?

smgogarten (approved):
Sorry for the delay; I was on vacation. Thanks for the bug fix!

PR #48, opened 2020-09-15
Matt branch

mconomos:
There are actually two things in here. One is the fix for the varComp name issue for rescaling variance. The second is some faster code for getting the unique set of sample.id in makeSparseMatrix (it was painfully slow before with a large data.frame).


smgogarten (approved):
I saw you copied code from another function, but the way it was originally written annoyed me so I made it prettier.

PR #49, opened 2020-09-16
add BinomiRare and CMP tests

smgogarten:


tamartsi:
Here is information to add about BinomiRare and CMP:

    The BinomiRare test, run by using \code{test = "BinomiRare"}, and the CMP test, run by using \code{test = "CMP"} are carrier-only, robust tests. Both tests focuse on carriers of the rare variant allele ("carriers"), and use the estimated probabilities of the binary outcome within the carriers, estimated under the null of not association, and the actual number of observed outcomes, to compute p-values. BinomiRare uses the Poisson-Binomial distribution, and the CMP uses the Conway-Maxwell-Poisson distribution, and is specifically designed for mixed models. These tests provide both a tranditional p-value (\code{"pval"}) and a mid-p-value (\code{"midp"}), which is less conservative/more liberal, with the difference being more pronounced for small number of carriers. The BinomiRare test is described in Sofer (2017) and compared to the Score and SPA in various settings in Sofer and Guo (2020). A reference to the CMP test and the BinomiRare for mixed models will be provided once a preprint is availabe.

tamartsi:
I think we should add the reporting of both the usual p-value and the mid-p-values for both BinomiRare and CMP test. Here's a function that returns both p-values for BinomiRare: 

.poibinP <- function(n.carrier, n.D.carrier, prob.vec){
  stopifnot(n.D.carrier <= n.carrier, length(prob.vec) == n.carrier)
  d.poibin <- poibin::dpoibin(0:n.carrier, prob.vec)
  prob.cur <- d.poibin[n.D.carrier + 1]
  pval <- prob.cur + sum(d.poibin[d.poibin < prob.cur])
  mid.pval <- 0.5*prob.cur + sum(d.poibin[d.poibin < prob.cur])
  
  pvals <- c(pval, mid.pval)
  names(pvals) <- c("pval", "mid.pval")
  return(pvals)
}

R/testGeno.R
```diff
@@ -35,52 +36,75 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
...
-        res <- .testGenoSingleVarScore(Gtilde, G, nullmod$resid, nullmod$RSS0)
         # saddle point approximation
         res <- SPA_pval(score.result = res, nullmod = nullmod, G = G, pval.thresh = recalc.pval.thresh)
     }
     
     if (test == "BinomiRare"){
-        if (nullmod$family$mixedmodel) stop("BinomiRare should be used for IID observations.")
-        if (nullmod$family$family != "binomial") stop("BinomiRare should be used for disease (binomial) outcomes.")
-    	res <- .testGenoSingleVarBR(nullmod$outcome, probs=nullmod$fitted.values, G)
+      if (nullmod$family$family != "binomial") stop("BinomiRare can only be used for binomial family.")
```
mconomos, line 44: @smgogarten should this check come earlier? before running the score test.

R/testGeno.R
```diff
@@ -35,52 +36,75 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
...
+      if (nullmod$family$mixedmodel) { ## if this is a mixed model, use conditional probabilities $
+        phat <- expit(nullmod$workingY - nullmod$resid.conditional)
+      } else{ ## not a mixed model
+        phat <- nullmod$fitted.values
+      }
+      score.pval <- if(calc.score) res$Score.pval else NULL
+      res <- .testGenoSingleVarBR(nullmod$outcome, probs=phat, G, score.pval=score.pval, pval.thresh=recalc.pval.thresh)
+    }
+
+    if (test == "CMP"){
```
mconomos, line 55: @tamartsi should there also be a check here for the null model being a binomial family?

PR #50, opened 2020-09-29
replace tidyverse with data.table

smgogarten:
Use data.table::rbindlist in place of both do.call(rbind) and dplyr::bind_rows, as it is [faster than both](https://rstudio-pubs-static.s3.amazonaws.com/406521_7fc7b6c1dc374e9b8860e15a699d8bb0.html).

With the removal of bind_rows, few other tidyverse dependencies remained. Replace dplyr with data.table solution in .matchAlleles, and replace simple dplyr code with base R in jointScoreTest.

With these changes, GENESIS depends on data.table but not tidyverse. Mixing both in the same package has been bothering me.


man/effectAllele.Rd
```diff
@@ -53,7 +53,7 @@ assoc <- assocTestSingle(iterator, nullmod)
 
 # add effect allele to the results
 eff <- effectAllele(seqData, variant.id=assoc$variant.id)
-assoc <- dplyr::left_join(assoc,eff)
+assoc <- merge(assoc, eff)
```
amstilp: I think `merge` doesn't preserve order while `left_join` does. Will this reorder `assoc`?  Does it matter?

amstilp:
The updates to the `jointScoreTest` function look good. I don't know the `data.table` package well enough to make any useful comments in those changes.

man/effectAllele.Rd
smgogarten: It won't in this case, because it's merging on variant.id and assoc is already in order by variant.id. This is just an example so I don't think it matters. In practice I'd want to use left_join, but I don't want to leave dplyr in Suggests for just this single example. I just had an idea, though - I could comment out that line so the dependency is not there but users can still see it.

smgogarten: Changing my mind yet again... we already have ggplot2 in Suggests so I might as well include dplyr there as well.

mconomos (approved):

PR #51, opened 2020-10-06
rename output of jointScoreTest to be consistent with assocTestSingle

smgogarten:
Joint tests are already present in GENESIS in the GxE option in
assocTestSingle. Set the names of jointScoreTest outputs to be
consistent.


R/jointScoreTest.R
```diff
@@ -51,20 +51,20 @@ jointScoreTest <- function(nullmod, G) {  # # Check rownames/colnames match.
   # Create fixed effects data frame.
   fixef <- data.frame(
     Est = as.vector(beta),
-    SE = se,
+    Est.SE = se,
```
mconomos: @smgogarten in the null model "fixef" output, we just call this SE. I'm guessing that's where @amstilp took the naming from. See the file prepareOutputForNullModel.R

Ultimately, it would be better if these were _all_ consistent, but I'm not sure which way we should go with this one if we don't want to change one of the older ones.

R/jointScoreTest.R
```diff
@@ -51,20 +51,20 @@ jointScoreTest <- function(nullmod, G) {  # # Check rownames/colnames match.
...
 
   res <- list()
 
-  res$Stat.joint <- as.numeric(Stat.joint)
-  res$pval.joint <- as.numeric(pval.joint)
-  res$pve.joint <- pve.joint
+  res$Joint.Stat <- as.numeric(Stat.joint)
+  res$Joint.pval <- as.numeric(pval.joint)
+  res$Joint.PVE <- pve.joint
   res$fixef <- fixef
```
mconomos, line 68: This change makes sense for consistency elsewhere.

amstilp (approved):

R/jointScoreTest.R
smgogarten: Good point, this should match the null model. I changed it back.

mconomos: To Adrienne's point, we could make them all be Est.SE. We do plan to refactor the null model output format in the near future (might make that a Q4 2020 SBG update). If we think it's best to make them all be Est.SE, then we should make the jointScoreTest one be that now. I have mixed feelings if that is the right way to go. I think the null model output was named SE to be like lm. In the assoc output, we have both a Score.SE and Est.SE, so we needed to specify which was which.

PR #52, opened 2020-10-10
Small null model

smgogarten:
BinomiRare and CMP don't require matrix objects in the null model, so provide an option to remove these to save memory. The `smallNullModel` function can be called independently, or at the end of `fitNullModel` is `return.small=TRUE`. In the latter case, we can also save time by skipping the call to `nullModelTestPrep`.


R/jointScoreTest.R
```diff
@@ -20,14 +20,14 @@ jointScoreTest <- function(nullmod, G) {  # # Check rownames/colnames match.
   }
 
   # Reorder samples.
-  idx <- match(rownames(nullmod$model.matrix), rownames(G))
+  idx <- match(rownames(null.model$model.matrix), rownames(G))
   G <- G[idx, , drop = FALSE]
 
   # Genotype adjusted for covariates and random effects.
-  Gtilde <- calcGtilde(nullmod, G)
+  Gtilde <- calcGtilde(null.model, G)
```
mconomos, line 32: @smgogarten This will fail with a "small" null model. Should there be any sort of check or warning?

R/varCompCI.R
```diff
@@ -1,30 +1,30 @@
-varCompCI <- function(nullMMobj, prop=TRUE){
+varCompCI <- function(null.model, prop=TRUE){
```
mconomos, line 1: thank you for changing this! haha

man/fitNullModel.Rd
```diff
@@ -52,6 +57,7 @@ nullModelInvNorm(null.model, cov.mat = NULL, norm.option = c("by.group", "all"),
   \item{max.iter}{The maximum number of iterations allowed to reach convergence.}
   \item{EM.iter}{The number of EM iterations to run prior to AIREML; default is 0.}
   \item{drop.zeros}{Logical indicator of whether variance component terms that converge to 0 should be removed from the model; the default is TRUE. See 'Details' for more information.}
+  \item{return.small}{Logical for whether to return a small version of the null model without NxN matrices. Default is FALSE; only set to TRUE for use in association tests with \code{test = "BinomiRare"} or \code{"CMP"} and \code{recalc.pval.thresh = 1}.}
```
mconomos: @smgogarten should this say `\code{test = "CMP"}` ?

man/fitNullModel.Rd
```diff
@@ -21,13 +22,15 @@
   association testing.
 
   \code{nullModelInvNorm} does an inverse normal transform of a previously fit null model.
+
+  \code{smallNullModel} returns a small version of the null model with no NxN matrices.
```
mconomos: A minor point, but should the names be either:
`smallNullModel` and `invNormNullModel`
or
`nullModelSmall` and `nullModelInvNorm`

Wondering if the symmetry is worthwhile. 

The first option is also symmetric with `fitNullModel`, but would require changing the existing `nullModelInvNorm` function name. Might not be the biggest deal as it's probably not used a ton?

R/fitNullModel_internal.R
```diff
@@ -89,8 +89,12 @@
...
 
     out.class <- class(out)
-    nullprep <- nullModelTestPrep(out)
-    out <- c(out, nullprep)
+    if (!return.small) {
+        nullprep <- nullModelTestPrep(out)
+        out <- c(out, nullprep)
+    } else {
+        out <- smallNullModel(out)
+    }
```
mconomos, line 96: I think we still need a couple of things that `nullModelTestPrep` does, even in the small case. Specifically, we need to calculate `resid` and `RSS0` (and maybe `Ytilde`).

Speaking of `Ytilde`, I do want to rename this to `resid.cholesky` at some point (to match `resid.marginal` and `resid.conditional`). Not sure if this would be a good time to do so.

R/jointScoreTest.R
amstilp: If we want to check if null models are "small", then maybe the null model should have a list element indicating if it's "small' or "full". This would be pretty quick and could be added when making other null model updates.

An alternative solution is turn null models into S4 Classes (a base class, and then extend that for small vs full null models, maybe). Then you could specify in the various signatures what you needed. I think that's the "best practices" way, but it would be a big change.

R/testGeno.R
```diff
@@ -12,6 +12,10 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
     test <- match.arg(test)
     calc.score <- test %in% c("Score", "Score.SPA") | (recalc.pval.thresh < 1)
 
+    if (is.null(nullmod$cholSigmaInv) && (calc.score || !is.null(E))) {
+        stop("small null model cannot be used with options provided")
+    }
+
```
amstilp: This should be a function (`isSmallNullModel`?), since it may end up being useful other places.

R/fitNullModel.R
```diff
@@ -176,3 +179,12 @@ nullModelInvNorm <- function(null.model, cov.mat = NULL,
...
 }
+
+
+## Return a small version of the null model (no NxN matrices)
+smallNullModel <- function(null.model) {
+    null.model$cholSigmaInv <- NULL
+    null.model$CX <- NULL
+    null.model$CXCXI <- NULL
+    null.model
+}
```
amstilp: Also buried in a comment responding to @mconomos's review:
- Should we add a list element indicating if the null model is small? then we can check for the existence of that list element instead of checking for all the required elements.

Alternatively, turn null models into S4 objects - a base class, and two classes that extend that base class (eg fullNullmodel and smallNullModel) - and use signatures. This is a much bigger chance.

R/fitNullModel_internal.R
amstilp: For renaming `Ytilde`: could be done in the general null model output updates, since the structure will completely change.

man/fitNullModel.Rd
smgogarten: `nullModelInvNorm` is used in the analysis pipeline, so it would require updating the null model script there. I think it might be better if it's not symmetric - `fitNullModel` takes other inputs and returns a null model, while `nullModelInvNorm` and `nullModelSmall` take a null model as input (and return a modified null model).

R/fitNullModel_internal.R
smgogarten: Where do `resid` and `RSS0` get used when the other components are not? I couldn't find anywhere, but I might be missing something.

R/testGeno.R
smgogarten: If we're changing `smallNullModel` to `nullModelSmall`, should we call this `isNullModelSmall`?

R/fitNullModel.R
smgogarten: Adding the `isNullModelSmall` function should solve this problem in a simple way. It doesn't even need to check for all the elements, just `cholSigmaInv`.

R/fitNullModel_internal.R
smgogarten: Is it a problem that running `nullModelSmall` on an existing null model preserves `resid`, `RSS0`, and `Ytilde`, but running `fitNullModel(return.small=TRUE)` returns a null model without them?

amstilp: From a user perspective, I think it would be confusing if there were different sets of list elements in the returned null model from `nullModelSmall` and `fitNullModel(return.small=TRUE)`. That said, the null model report in the analysis pipeline needs `Ytilde` and `resid`.

smgogarten: In that case, maybe we should just always run `nullModelTestPrep`, and only remove the NxN matrices?

mconomos: I realize now that the use case I was thinking of is a future use case. Once I implement the "fast variance approximation for score tests" that SAIGE uses, then you would use a small null model to do assocTestSingle (maybe also aggregate?), but it would require the `resid` and `RSS0` elements. 

It may make the most sense to run `nullModelTestPrep` and then remove the NxN matrices. Using `return.small = TRUE` wouldn't save any time, but it would save storage space in the resulting object.

Another possibility would be to make `nullModelTestPrep` a separate step / function to be called on a small null model to make it "full", but that would be a bigger change for users and likely doesn't make as much sense.

R/testGeno.R
mconomos: `isNullModelSmall` seems reasonable to me, and makes me laugh. However, I think the check is simply `is.null(nullmod$cholSigmaInv)`.

R/fitNullModel_internal.R
mconomos: > For renaming `Ytilde`: could be done in the general null model output updates, since the structure will completely change.

Fine with me to wait til we do this.

PR #53, opened 2020-10-13
Add error with collinear covariates

amstilp:
* Add an error in `createDesignMatrix` when some combination of covariates are collinear
* Add test to verify that error is produced when appropriate
* Update man page for `fitNullModel` to reflect this new error

### DIFF
--- a/man/fitNullModel.Rd
+++ b/man/fitNullModel.Rd
@@ -63,7 +63,7 @@ nullModelInvNorm(null.model, cov.mat = NULL, norm.option = c("by.group", "all"),
 \details{
     If \code{x} is a data.frame, the rownames of \code{x} must match the row and column names of \code{cov.mat} (if \code{cov.mat} is specified). If \code{x} is an \code{\link{AnnotatedDataFrame}} or other object containing an \code{\link{AnnotatedDataFrame}}, \code{x} will be re-ordered (if necessary) so that \code{sample.id} or \code{scanID} is in the same order as the row and column names of \code{cov.mat}.
 
-    The code checks for multicollinearity of the design matrix by checking that its rank is equal to the number of columns; if the rank is smaller, it fails with an error.
+    The code checks for multicollinearity of covariates by checking that the rank of the design matrix is equal to the number of columns; if the rank is smaller, it fails with an error.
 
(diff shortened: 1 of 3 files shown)


man/fitNullModel.Rd
```diff
@@ -63,19 +63,21 @@ nullModelInvNorm(null.model, cov.mat = NULL, norm.option = c("by.group", "all"),
 \details{
     If \code{x} is a data.frame, the rownames of \code{x} must match the row and column names of \code{cov.mat} (if \code{cov.mat} is specified). If \code{x} is an \code{\link{AnnotatedDataFrame}} or other object containing an \code{\link{AnnotatedDataFrame}}, \code{x} will be re-ordered (if necessary) so that \code{sample.id} or \code{scanID} is in the same order as the row and column names of \code{cov.mat}.
 
+    The code checks for collinear of the design matrix by checking that its rank is equal to the number of columns; if the rank is smaller, it fails with an error.
+
```
mconomos, line 67: Suggested slight rewording:

The code checks for multicollinearity of covariates by checking that the rank of the design matrix is equal to the number of columns; if the rank is smaller, it fails with an error.

tests/testthat/test_nullModel.R
```diff
@@ -28,6 +28,28 @@ test_that("design matrix with missing reference level", {
...
+    # No failure with no covariates.
+     expect_silent(nm <- fitNullModel(dat, outcome="a"))
+     # No failure with no colinear covariates.
+     expect_silent(fitNullModel(dat, outcome="a", covars = c("b", "c", "d")))
+     # Simple case where one covariate equals another.
+     expect_error(fitNullModel(dat, outcome="a", covars = c("b", "c", "d", "e")),
+                  "collinear")
+     # Slightly more complicated case involving linear combination of more than one covariate.
+     expect_error(fitNullModel(dat, outcome="a", covars = c("b", "c", "d", "f")),
+                  "collinear")
```
amstilp: I decided to test the behavior of `fitNullmodel`, which is the exported, instead of `createDesignMatrix`, where the actual check is implemented -- following https://r-pkgs.org/tests.html which says "Focus on testing the external interface to your functions". The tests also pass when calling `createDesignMatrix` instead, so I can change it if desired.

man/fitNullModel.Rd
amstilp: Fixed, thanks

mconomos (approved):

smgogarten (approved):

PR #54, opened 2020-10-14
change small.samp.correct default to TRUE

smgogarten:
@mconomos said:
It used to be this way and we changed it at one point. PC-Relate is one of the main things I get email questions about, and it's often from people working with small data sets. I often end up telling them that they should set that to TRUE. The code already has a check built in to set it back to FALSE if someone attempts to use it with parameters for which it won't work.


man/pcrelate.Rd
```diff
@@ -68,7 +68,7 @@ correctK0(kinBtwn)
...
   \item{maf.bound.method}{How individual-specific minor allele frequency estimates less that \code{maf.thresh} are handled. When set to `filter` (default), SNPs for which an individual's estimated individual-specific minor allele frequency are below \code{maf.thresh} are excluded from the analysis for that individual. When set to `truncate`, estimated individual-specific minor allele frequncies below \code{maf.thresh} have their value set to \code{maf.thresh}.}
- \item{small.samp.correct}{Logical indicator of whether to implement a small sample correction.}
+ \item{small.samp.correct}{Logical indicator of whether to implement a small sample correction. The default is \code{TRUE}, but must be set to \code{FALSE} if \code{sample.block.size} is less than the number of samples or if \code{scale} is \code{'none'}.}
```
smgogarten, line 71: @mconomos should we add any additional instructions for when it's advised to set it to `FALSE`?

mconomos (approved):
Thanks!

man/pcrelate.Rd
mconomos: It will automatically switch to false when the number of sample blocks is greater than 1 (and print a message). We could say something like "it's computationally demanding on large samples, so we don't recommend it if your sample size is > 10,000(?)"

I think we _did_ use it for OLGA, which was like 12,000, but probably less necessary.

mconomos: (sorry i missed this question before you merged)

PR #57, opened 2020-11-06
Restructure null model

amstilp:
This is a major change in restructuring the null model to include a data frame of sample-level values called `nullmod$fit`. This data frame includes the following columns:
* outcome
* workingY
* fitted.values
* resid.marginal
* resid.conditional
* resid.cholesky (renamed; used to be Ytilde)
* resid
* sample.id (if called using an AnnotatedDataFrame)

This branch also adds elements to track the null model string: `nullmod$model`, `nullmod$outcome`, and `nullmod$covars`. I've added some comments about how I decided to implementing this change to this pull request.

I'm also considering adding a function to update older null models to the new format, but haven't decided if I have time to do it yet.


R/fitNullModel.R
```diff
@@ -17,25 +17,36 @@ setMethod("fitNullModel",
...
                             drop.zeros=drop.zeros,
                             return.small=return.small,
                             verbose=verbose)
+              rownames(out$fit) <- rownames(desmat$y)
+
+              # Add model string elements here because we need the outcome string.
+              out$model <- .modelString(outcome, covars = covars, random = names(cov.mat),
+                                        group.var = group.var, inverse_normal = FALSE)
+              out$outcome <- .modelOutcomeString(outcome, inverse_normal = FALSE)
+              out$covars <- covars
```
amstilp: I ended up deciding to add this to the `fitNullModel` wrapper function instead of in any of the sub-functions for a couple reasons. One side effect is that the model will still include any dropped covariates. Here are my reasons:
* I had first considered adding it to `.fitNullModel`, which accepts the design matrix as an argument. In the case of categorical covariates, this design matrix has multiple indicator variables for that covariate (eg instead of `study`, it will have `studyB`, `studyC`, etc.). If you tried to use the colnames of the design matrix to get the list of covariates, you would end up with `y ~ studyB + studyC` instead of `y ~ study`.
* It would be possible to have `createDesignMatrix` return the list of covariates actually included in the model and then use that list element to create the model string. In this case, we could remove dropped covariates from the model, but it was faster to code it this way. It is also more consistent with `lm` to keep dropped covariates in the model string.

R/fitNullModel.R
```diff
@@ -193,3 +205,40 @@ nullModelSmall <- function(null.model) {
...
+  # What about the case where one element of the matrix has names, and the others don't?
+  if (is.null(names(cov.mat))) {
+    names(cov.mat) <- LETTERS[1:length(cov.mat)]
+  } else {
+    if (any(names(cov.mat) == "")) {
+      stop("Some names for cov.mat list are missing.")
+    }
+  }
+  cov.mat
+}
```
amstilp: To create the null model string, we need to make sure the covariance matrices have consistent names throughout the code (e.g. in the model string and in the `varComp` null model element), but the names are assigned in different places. This function's purpose is to standardize naming across multiple functions.

R/updateNullModCond.R
```diff
@@ -1,37 +1,40 @@
...
+    ## a few checks that may be transfered to wrapper function:
     if (nullmod$hetResid & is.null(nullmod$group.idx)) stop("group indices are required for updating the null model")
 
     #if (updateVarComp){ ## this check may be pulled out for wrapper function.
     if (is.null(covMatList)) stop("covMatList is needed for udpating variance components")
     if (nullmod$hetResid & is.null(nullmod$group.idx)) stop("group indices are required for updating variance components")
     #}
 
-    
+    if (is.null(colnames(G))) colnames(G) <- paste0("var_", 1:ncol(G))
```
amstilp, line 17: To add conditional variants to the model string, we need to make sure the genotype matrix has column names for the variant. If `G` has colnames, it will use those. If it does not, I used the convention from analysis_pipeline where they start with "var_", but because the variant.id is unknown, they are just numbered sequentially (e.g., var_1, var_2, var_3, etc.).

R/updateNullModCond.R
```diff
@@ -1,37 +1,40 @@
...
+
     ## add any extra slots
     extra <- setdiff(names(nullmod), names(new.nullmod))
     new.nullmod <- c(new.nullmod, nullmod[extra])
-    return(new.nullmod)  
+    # Update the model string.
+    previous_covar_string <- .modelCovarString(new.nullmod$covars)
+    new.nullmod$covars <- c(new.nullmod$covars, colnames(G))
+    new_covar_string <- .modelCovarString(new.nullmod$covars)
+    new.nullmod$model <- sub(previous_covar_string, new_covar_string, new.nullmod$model)
```
amstilp: The model string is added in the `fitNullModel` wrapper, but this function only calls the internal `.fitNullModel` function so it has to jump through some hoops to update the model string.

R/updateNullModOutcome.R
```diff
@@ -65,57 +63,64 @@ updateNullModOutcome <- function(nullmod,
...
     ## add any extra slots
     extra <- setdiff(names(nullmod), names(new.nullmod))
     new.nullmod <- c(new.nullmod, nullmod[extra])
+    # Add sample id.
+    new.nullmod$fit$sample.id <- nullmod$fit$sample.id
+    # Update model string (but keep outcome the same?).
+    previous_model <- new.nullmod$model
+    # What happens if covMatList is different than what was passed to the original null model call?
+    new.nullmod$model <- paste(sprintf("rankInvNorm(resid(%s))", new.nullmod$outcome), "~",
+                               strsplit(previous_model, " ~ ")[[1]][2])
```
amstilp: The model string will be incorrect if the user passes a different set of covariance matrices when inverse-normal transforming the null model, but I can't see a valid reason that someone would do that (other than user error).

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+
+The \code{fit} data frame contains the following columns, depending on which type of model was fit:
+\describe{
+  \item{outcome}{The original outcome vector. When \code{family} is gaussian, this is equal to \code{workingY}.}
+  \item{workingY}{The "working" outcome vector. When \code{family} is gaussian, this is just the original outcome vector. When \code{family} is not gaussian, this is the PQL linearization of the outcome vector. This is used by \code{\link{assocTestSingle}} or \code{\link{assocTestAggregate}} for genetic association testing. See 'Details' for more information.}
+  \item{fitted.values}{The fitted values from the model; i.e. W*beta where W is the design matrix and beta are the effect size estimates for the fixed effects.}
+  \item{resid.marginal}{The marginal residuals from the model; i.e. Y - W*beta where Y is the vector of outcome values.}
+  \item{resid.conditional}{The conditional residuals from the model; i.e. Y - W*beta - Z*u.}
+  \item{resid.cholesky}{The Cholesky residuals from the model; i.e. - XXX}
+  \item{resid}{XXX. These are used in the final score tests for association.}
```
amstilp: I wasn't sure how to describe these two variables in the documentation. @mconomos do you have suggestions?

tests/testthat/test_nullModel.R
```diff
@@ -317,3 +359,76 @@ test_that("data.tables are supported", {
...
+  expect_equal(nullmod_update$covars, nullmod_geno$covars)
+
+  # Genotype matrix with colnames.
+  colnames(G) <- "variant"
+  nullmod_update <- updateNullModCond(nullmod, G, covMatList=covMat, verbose=FALSE)
+  expect_equal(nullmod_update$model, "a ~ b + variant + (1|A)")
+  expect_equal(nullmod_update$covars, c("b", "variant"))
+
+})
+#}
```
amstilp: This test was not run due to the `if(FALSE)` statement, and it in a different test_* source file. I rewrote it a bit and moved it to the test_nullModel source file. Is there a reason to not run it?

R/updateNullModCond.R
smgogarten: This function is not currently exported, and is awaiting a wrapper of its own. Like `updateNullModelOutcome`, which is called by `nullModelInvNorm`, it was intended to be an internal function. So would it make sense to handle this part in a future wrapper?

R/updateNullModOutcome.R
smgogarten: See above comment - if model strings are usually handled in wrapper vs internal functions, would it make sense to move this to `nullModelInvNorm`?

smgogarten (changes requested):
For backwards compatibility, I think we should have a function that converts from old to new format, at least for the parts of the new format that are breaking changes (so just renaming list elements, but don't bother creating the model string). We might not even need to export this function, but just call it at the beginning of any other function that takes a null model object, and have it print a warning about the format change (e.g., "this null model was created with an older version of GENESIS and is being updated to use the current version. please consider re-running the null model with the current version.") It will probably be less work than answering lots of questions about "why is my null model mot working anymore?"

R/updateNullModCond.R
amstilp: Yes, I agree that handling it in a wrapper is better. I could add a comment to this part saying that these lines should be moved to a wrapper once it's written. If we remove them entirely, then whoever wrote the wrapper would have to figure out how to code it again. Thoughts?

R/updateNullModOutcome.R
amstilp: Agree, same comment as above.

R/nullModelTestPrep.R
```diff
@@ -30,16 +30,17 @@ nullModelTestPrep <- function(nullmod){
     # compute residual sum of squares under the null model
     RSS0 <- as.numeric(crossprod(Ytilde))
 
-    return(list(Ytilde = Ytilde, resid = resid, CX = CX, CXCXI = CXCXI, RSS0 = RSS0))
-    # return(list(Ytilde = Ytilde, resid = resid, CX = CX, CXCXI = CXCXI, qr = qrmod))
+    #return(list(Ytilde = Ytilde, resid = resid, ))
+    out <- list(resid.cholesky = Ytilde, resid = resid,
```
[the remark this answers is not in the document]
amstilp: I like this! @mconomos do you agree with changing `resid` to `resid.PY`?

amstilp:
@smgogarten I can add a function to convert a null model in the old format to the new format. I'd like to have tests for it, though. I can think of these options for approaching the testing:
1. Create null models using the older version of the code and store them as data for testing somewhere. This is more robust but would lead to a lot of small test data files.
2. Write a function to create a null model in the old format. This is less robust because there are a lot of slightly different formats depending on which type of null model you ran (logistic, linear, linear mixed, etc.) and how you called it (with an Annotated Data Frame vs. with a data frame with rownames).

I'm leaning towards 1) but let me know if you think differently.

R/updateNullModCond.R
smgogarten: Leaving it in with a comment sounds like a good plan.

tests/testthat/test_nullModel.R
smgogarten: I think it was giving errors in R CMD check that I did not understand (maybe only on the Bioconductor build system), and since we don't use or export this function anyway, I just stopped running the test as a temporary fix. If it works after rewriting, it should be run.

smgogarten:
I think storing test data files is fine. We should put them in `inst/extdata` to avoid having to document them, as everything in `data` requires a man page.

amstilp:
I think we can keep test data in the `tests/` directory (see [Hadley Wickham's comment](https://r-pkgs.org/data.html#other-data)). That way they're less exposed to users who might think they're useful (when they really are not, outside of the tests). I'll see if it works that way.

R/updateNullModOutcome.R
amstilp: Oops, realized this one did have a wrapper - I ended up moving it to the wrapper.

R/updateNullModCond.R
amstilp: I added a comment above these lines.

R/updateNullModOutcome.R
amstilp: I moved the model string update lines to the wrapper.

R/prepareOutputForNullModel.R
```diff
@@ -29,10 +29,18 @@
     zeroFLAG <- NULL
     RSS <- ifelse(family$family == "gaussian", sum(resid.marginal^2)/varComp/(nrow(X) - ncol(X)), 1)
 
+    # Sample-level data frame.
+    fit <- data.frame(
+      outcome = as.vector(y),
+      workingY = as.vector(workingY),
+      fitted.values = unname(fitted.values),
+      resid.marginal = unname(resid.marginal)
+    )
```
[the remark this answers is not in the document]
amstilp: Thanks, good catch. I think we want `stringsAsFactors=FALSE` in the fit data frames everywhere.

man/fitNullModel.Rd
amstilp: I updated this from commented out lines from the previous help file:
```
  \item{resid.cholesky}{phenotype values adjusted for covariates and random effects. This adjusted phenotype vector approximately follow a distribution with mean 0 and an identity covariance matrix. Linear regression of this adjusted phenotype vector on an equivalently adjusted genotype vector provides the same estimates as fitting the full GLS model (by the Frisch-Waugh-Lovell theorem).}
  \item{resid}{phenotype values pre-multiplied by the projection matrix that adjusts for covariates and random effects. These projected phenotype values are essentially what are correlated with genotype values for association testing.}
```

R/nullModelTestPrep.R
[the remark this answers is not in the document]
amstilp: I renamed this column to `resid.PY`.

tests/testthat/test_nullModel.R
amstilp: It runs in R CMD check - we can take it out again if it fails in the Bioconductor build system and we don't know how to fix it.

amstilp:
Ok, I think I've made all of @smgogarten's suggested changes. Anything else before updating the version and NEWS, and merging?

R/fitNullModel.R
```diff
@@ -193,3 +221,83 @@ nullModelSmall <- function(null.model) {
 isNullModelSmall <- function(null.model) {
     is.null(null.model$cholSigmaInv)
 }
+
+
```
mconomos, line 227: @amstilp This probably needs to be updated for GxE tests. The string is more complicated I think.

R/fitNullModel.R
```diff
@@ -193,3 +221,83 @@ nullModelSmall <- function(null.model) {
...
+  # Update.
+  nullmod$fit <- data.frame(
+    outcome = as.vector(nullmod$outcome),
+    workingY = as.vector(nullmod$workingY),
+    fitted.values = as.vector(nullmod$fitted.values),
+    resid.marginal = as.vector(nullmod$resid.marginal),
+    resid.PY = as.vector(nullmod$resid),
+    resid.cholesky = as.vector(nullmod$Ytilde),
+    stringsAsFactors = FALSE
+  )
```
mconomos, line 285: @amstilp I think you're missing `resid.conditional`

R/nullModelSubset.R
```diff
@@ -1,19 +1,17 @@
 ## idx.exclude are indices of individuals that should be excluded (e.g. because of missing genotypes)
 nullModelSubset <- function(nullmod, idx.exclude){
-    for (v in c("sample.id", "fitted.values", "resid.marginal", "resid.condition", "workingY", "outcome")) {
-        nullmod[[v]] <- nullmod[[v]][-idx.exclude]
-    }
+    nullmod$fit <- nullmod$fix[-idx.exclude, ]
```
mconomos: @amstilp typo "fix" instead of "fit"

R/testGeno.R
```diff
@@ -34,36 +34,36 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
     if(calc.score){
         Gtilde <- calcGtilde(nullmod, G)
         if(is.null(nullmod$RSS0)){
```
mconomos: @amstilp this RSS0 calculation could actually move into your "update old null models" function I think. This is also a fix for old versions.

R/testGeno.R
```diff
@@ -148,21 +148,21 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
     E <- as.matrix(E)
     p <- ncol(G)
     v <- ncol(E) + 1
-    n <- length(nullmod$Ytilde)
+    n <- length(nullmod$fit$resid.cholesky)
```
mconomos: could just make this `nrow(nullmod$fit)`

R/variantSetTests.R
```diff
@@ -58,14 +58,14 @@ testVariantSet <- function( nullmod, G, weights,
     }else{
         burden <- colSums(t(G) * weights)
     }
-    
+
     # adjust burden for covariates and random effects
     if (burden.test == "Score") {
         Gtilde <- calcGtilde(nullmod, burden)
         if(is.null(nullmod$RSS0)){
```
mconomos: @amstilp again - i think this could be moved to your update old null models function

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
     \item{converged}{A logical indicator of whether the AIREML procedure for estimating the random effects variance components converged.}
     \item{zeroFLAG}{A vector of logicals the same length as \code{varComp} specifying whether the corresponding variance component estimate was set to 0 by the function due to convergence to the boundary in the AIREML procedure.}
     \item{RSS}{The residual sum of squares from the model fit. When \code{family} is gaussian, this will typically be 1 since the residual variance component is estimated separately.}
-    \item{sample.id}{A vector of IDs for the samples used in the
-      analysis.}
+    \item{model}{The model string.}
+    \item{outcome}{The outcome variable name.}
```
mconomos: @amstilp is there any reason to be concerned that `outcome` used to be the actual outcome vector, and now it's just a string for the outcome variable?

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+    \item{covars}{A vector of covariate names}
+}
     %\item{resid}{phenotype values pre-multiplied by the projection matrix that adjusts for covariates and random effects. These projected phenotype values are essentially what are correlated with genotype values for association testing.}
     %\item{Ytilde}{phenotype values adjusted for covariates and random effects. This adjusted phenotype vector approximately follow a distribution with mean 0 and an identity covariance matrix. Linear regression of this adjusted phenotype vector on an equivalently adjusted genotype vector provides the same estimates as fitting the full GLS model (by the Frisch-Waugh-Lovell theorem).}
     %\item{RSS0}{the sum of the `Ytilde` values squared. This is the sum of squared residuals under the null hypothesis of no genetic effect for the covariate and random effect adjusted model using the Frisch-Waugh-Lovell theorem.}
+
+The \code{fit} data frame contains the following columns, depending on which type of model was fit:
+\describe{
+  \item{outcome}{The original outcome vector. When \code{family} is gaussian, this is equal to \code{workingY}.}
```
mconomos: Probably don't need to say this is the same as `workingY`. That's covered in the `workingY` description.

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+  \item{workingY}{The "working" outcome vector. When \code{family} is gaussian, this is just the original outcome vector. When \code{family} is not gaussian, this is the PQL linearization of the outcome vector. This is used by \code{\link{assocTestSingle}} or \code{\link{assocTestAggregate}} for genetic association testing. See 'Details' for more information.}
+  \item{fitted.values}{The fitted values from the model; i.e. W*beta where W is the design matrix and beta are the effect size estimates for the fixed effects.}
```
mconomos: @amstilp could we change the `W` in these descriptions to `X`? I think it would be more consistent with the code; also, we often use `W` to describe a weight matrix in other places.

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+  \item{resid.marginal}{The marginal residuals from the model; i.e. Y - W*beta where Y is the vector of outcome values.}
+  \item{resid.conditional}{The conditional residuals from the model; i.e. Y - W*beta - Z*u.}
```
mconomos: , where Z is the design matrix and u are the effect size estimates for the random effects. Zu represents the effects captured by the random effects specified with the `cov.mat` input.

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+  \item{resid.cholesky}{phenotype values adjusted for covariates and random effects. This adjusted phenotype vector approximately follow a distribution with mean 0 and an identity covariance matrix. Linear regression of this adjusted phenotype vector on an equivalently adjusted genotype vector provides the same estimates as fitting the full GLS model (by the Frisch-Waugh-Lovell theorem).}
```
mconomos: @amstilp please replace with this:

{The Cholesky residuals from the model; a transformation of the marginal residuals using the estimated model covariance structure such that they are uncorrelated and follow a standard multivariate Normal distribution with mean 0 and identity covariance matrix asymptotically. Linear regression of the Cholesky residual vector on an equivalently transformed genotype vector provides the same estimates as fitting the full GLS model (by the Frisch-Waugh-Lovell theorem).}

man/fitNullModel.Rd
```diff
@@ -112,11 +111,25 @@ isNullModelSmall(null.model)
...
+  \item{resid.PY}{phenotype values pre-multiplied by the projection matrix that adjusts for covariates and random effects. These projected phenotype values are essentially what are correlated with genotype values for association testing.}
```
mconomos: @amstilp please replace with this:

{The outcome vector (Y) pre-multiplied by a projection matrix (P) that adjusts for covariates, random effects, and model covariance structure. These projected outcome values are essentially what are correlated with genotype values for association testing.

R/fitNullModel.R
```diff
@@ -193,3 +221,83 @@ nullModelSmall <- function(null.model) {
...
+    return(sprintf("rankInvNorm(resid(%s))", outcome))
+  } else {
+    return(outcome)
+  }
+}
+
+.setCovMatNames <- function(cov.mat) {
+  if (!is.list(cov.mat)) {
+    cov.mat <- list(cov.mat)
+  }
```
mconomos, line 252: I think that `.covMatNames` turns a list of length 1 into a matrix, then this function turns it back into a list. That might not be a problem, but I wonder if we can/should clean up all this covMat manipulation code.

R/fitNullModel.R
```diff
@@ -193,3 +221,83 @@ nullModelSmall <- function(null.model) {
...
+  # What about the case where one element of the matrix has names, and the others don't?
```
mconomos, line 253: Can we assign `LETTERS` to ones that have no name? Or should we just always require names?

R/fitNullModel.R
mconomos: I think it makes sense to have this here rather than in `.fitNullModel`. I still have a slight preference for dropped covariates to be removed from the model string (but could be convinced otherwise). It seems like `createDesignMatrix` could even be a good place to get that list of covariates. One tricky thing is that the code there can (reasonably) end up dropping some categories in a factor. Maybe that complication makes it worth just leaving as is.

R/fitNullModel.R
```diff
@@ -71,9 +82,10 @@ setMethod("fitNullModel",
                   ind <- match(.covMatNames(cov.mat), rownames(x))
                   x <- x[ind,]
               }
-              
+
               nm <- fitNullModel(x, outcome, covars, cov.mat, group.var, ...)
-              nm$sample.id <- rownames(nm$model.matrix)
+              # XXX: Decide if we should reorder columns in the fit data frame such that sample.id is the first column.
+              nm$fit$sample.id <- rownames(nm$model.matrix)
```
mconomos, line 88: When we use an AnnotatedDataFrame, are we going to end up with the sample.id being both a column and the rownames of the `fit` object?

R/fitNullModel.R
```diff
@@ -101,14 +113,30 @@ setMethod("fitNullModel",
...
+    # Update null model format.
+    null.model <- .updateNullModelFormat(null.model)
+
+    new.nullmod <- updateNullModOutcome(null.model, covMatList=cov.mat, rankNorm.option=norm.option,
+                         rescale=rescale, AIREML.tol=AIREML.tol,
                          max.iter=max.iter, EM.iter=EM.iter,
                          verbose=verbose)
+
+    # Add sample id.
+    new.nullmod$fit$sample.id <- null.model$fit$sample.id
```
mconomos, line 131: What if it's a nullmod where the `sample.id` are the rownames and not a column?

R/nullModelTestPrep.R
mconomos: call this `resid.PY` here for consistency in the code?

R/fitNullModel_internal.R
```diff
@@ -44,58 +44,61 @@
...
+            }
         } else{
             out <- .nullModOutReg(y, X, mod, family)
         }
     }
 
     out.class <- class(out)
     nullprep <- nullModelTestPrep(out)
-    out <- c(out, nullprep)
+    # Add to fit data frame.
```
mconomos, line 91: In line 16 above (can't leave a comment there), I think you could use your new covMat naming function

R/prepareOutputForNullModel.R
```diff
@@ -29,10 +29,19 @@
     zeroFLAG <- NULL
     RSS <- ifelse(family$family == "gaussian", sum(resid.marginal^2)/varComp/(nrow(X) - ncol(X)), 1)
 
+    # Sample-level data frame.
+    fit <- data.frame(
+      outcome = as.vector(y),
```
mconomos, line 34: Is it weird that we mix `outcome` and `workingY`? Should it be `Y` and `workingY` or `outcome` and `workingOutcome`? Maybe I'm overthinking.

R/testGeno.R
```diff
@@ -34,36 +34,36 @@ testGenoSingleVar <- function(nullmod, G, E = NULL, test = c("Score", "Score.SPA
     if(calc.score){
         Gtilde <- calcGtilde(nullmod, G)
         if(is.null(nullmod$RSS0)){
-            nullmod$RSS0 <- as.numeric(crossprod(nullmod$Ytilde))
+            nullmod$RSS0 <- as.numeric(crossprod(nullmod$fitresid.cholesky))
```
mconomos: typo: missing `$` between `fit` and `resid.cholesky`

mconomos (commented):
@amstilp - I spent some time trying to look at this in detail. I left a bunch of comments, let me know if you want to discuss any of them. It looks good though. Thanks!

R/fitNullModel.R
amstilp: I think the *null model* string should be the same though. The actual model strings are (ignoring random effects):
- null model string: y ~ x1 + x2 + x3
- regular test: y ~ x1 + x2 + x3 + snp
- GxE: y ~ x1 + x2 + x3 + snp + x1*snp

but in all cases, the null model string is just `y ~ x1 + x2 + x3`. You also use the same null model object to run the regular test as well as the GxE test, right?

R/nullModelSubset.R
amstilp: Fixed

R/fitNullModel.R
mconomos: Oh. You are absolutely correct. Sorry about that.

R/testGeno.R
amstilp: Good point -- done.

man/fitNullModel.Rd
amstilp: I thought about this briefly and decided not to worry about it, but I could be convinced. The only alternative I could think of to `outcome` would be `outcome_string`, but then we should change the other elements to `model_string` and `covar_strings` (which isn't really a string; it's a list of strings). So I stuck with `outcome` and figured that users would read the help or the `NEWS` if they noticed a difference.

man/fitNullModel.Rd
amstilp: Fixed

man/fitNullModel.Rd
amstilp: No problem.

man/fitNullModel.Rd
amstilp: Zu should be Z*u?

R/fitNullModel.R
amstilp: I decided against this because if I had a list like:
```
l <- list(
  A = matrix(runif(4), 2,2),
  matrix(runif(4), 2, 2)
)
```
It would assign `A` to the second list element unless we then also checked that no names were duplicated (which is probably a check we should do). I think it's better to have them all named by the user or non named by the user and then named by the code.

man/fitNullModel.Rd
mconomos: Yup.

R/fitNullModel.R
amstilp: Right now, yes. I could remove the rownames when adding the `sample.id` column without breaking anything, I think.

R/fitNullModel.R
amstilp: In that case, the `sample.id` column won't exist even after running this command. Because `null.model$fit$sample.id` is `NULL`, it sets `new.nullmod$fit$sample.id` to `NULL`, which means it's not in the new `fit` data frame. I'll add a comment explaining this.

amstilp: `rownames` are already set by the `updateNullModOutcome` function, because it calls `fitNullModel`.

R/fitNullModel.R
amstilp: Also noting if we removed the rownames when `sample.id` was present, we'd have to do it in a few wrappers (or write an internal function to add `sample.id` to the null model fit data frame, and then remove the rownames).

man/fitNullModel.Rd
mconomos: I was trying to look at what other functions do (e.g. `lm`, `lmer`) if anything. Had another idea that I thought I'd run by you. What if we had an object in the output called `model` that is a list with the elements `formula` (currently model), `outcome`, and `covars`. Perhaps we would also move `family` and `hetResid` into that list. Basically all the things that describe the model and aren't data. What do you think? Better or worse?

man/fitNullModel.Rd
mconomos: On line 120 above, I just noticed that the description of `RSS0` is commented out, but I'm not sure why.

R/fitNullModel.R
mconomos: Makes sense.

R/fitNullModel.R
mconomos: It probably doesn't really matter. If we want it to be "cleaner", the last thing you said seems reasonable. We could just have a function that operates at the end of fitting the null model that adds a `sample.id` column and removes the rownames if needed. Then the `fit` matrix would always look the same regardless of input data type, which might be nice. Your call.

R/fitNullModel.R
```diff
@@ -17,25 +17,36 @@ setMethod("fitNullModel",
...
+
+
+              out <- .fitNullModel(y=desmat$y, X=desmat$X, covMatList=cov.mat,
                             group.idx=desmat$group.idx, family=family,
                             start=start, AIREML.tol=AIREML.tol,
                             max.iter=max.iter, EM.iter=EM.iter,
                             drop.zeros=drop.zeros,
                             return.small=return.small,
                             verbose=verbose)
+              rownames(out$fit) <- rownames(desmat$y)
```
mconomos, line 40: What if we just assigned this to a column called `sample.id` here instead of the rownames? Sorry if I'm forgetting a reason why that doesn't work.

R/fitNullModel.R
```diff
@@ -193,3 +221,89 @@ nullModelSmall <- function(null.model) {
...
+    nullmod$workingY <- NULL
+    nullmod$fitted.values <- NULL
+    nullmod$resid.marginal <- NULL
+    nullmod$resid <- NULL
+    nullmod$Ytilde <- NULL
+  }
+
+  # Add RSS0
+  if (is.null(nullmod$RSS0)) {
+    warning(msg)
```
mconomos: will we get the same warning message twice? If a model doesn't have `RSS0`, it's also not going to have `fit`. Unless the user did something really weird and modified the output.

man/fitNullModel.Rd
amstilp: I think that makes sense.

man/fitNullModel.Rd
amstilp: It was like that when I started the changes. I can uncomment it.

R/fitNullModel.R
amstilp: I think @smgogarten said on Slack that the differences with `sample.id` were intentional for different input types (AnnotatedDataFrame vs. data.frame), so I tried to it keep it that way. I'm leaning towards just keeping it as is.

R/fitNullModel_internal.R
amstilp: Thanks, I missed that one.

R/fitNullModel.R
amstilp: See above, this was intentional behavior for data frames.

R/fitNullModel.R
amstilp: Yes, right now you get the warning twice. There was a version of the code that has `RSS0` but not the `fit` data frame, so sometimes it has to add both `fit` and `RSS0` and other times it just has to add `RSS0`. Maybe it's not worth printing the warning for missing `RSS0`. Thoughts?

R/prepareOutputForNullModel.R
amstilp: I think either keep it as is or change `workingY` to `workingOutcome`. The `workingY` element is used in a lot of places right now.

amstilp: Leaning towards keeping as is unless it's too confusing just based on time.

R/fitNullModel.R
mconomos: No strong opinion. It's fine to get the warning twice, or not bother for the RSS0 change, since it's such a minor change. Or check for either condition and then print the warning once.

man/fitNullModel.Rd
amstilp: I made this change, including moving `hetResid` and `family` to the new `model` list element.

R/fitNullModel.R
amstilp: I ended up taking out the warning in the RSS0 check.

man/fitNullModel.Rd
```diff
@@ -112,9 +111,6 @@ isNullModelSmall(null.model)
     \item{zeroFLAG}{A vector of logicals the same length as \code{varComp} specifying whether the corresponding variance component estimate was set to 0 by the function due to convergence to the boundary in the AIREML procedure.}
     \item{RSS}{The residual sum of squares from the model fit. When \code{family} is gaussian, this will typically be 1 since the residual variance component is estimated separately.}
     \item{RSS0}{the sum of the `Ytilde` values squared. This is the sum of squared residuals under the null hypothesis of no genetic effect for the covariate and random effect adjusted model using the Frisch-Waugh-Lovell theorem.}
```
mconomos: @amstilp we need to update the description of `RSS0`. 

the sum of `resid.cholesky` squared. This is the sum of squared residuals under the null hypothesis of no genetic effect for the covariate and random effect adjusted model using the Frisch-Waugh-Lovell theorem.

man/fitNullModel.Rd
```diff
@@ -112,9 +111,6 @@ isNullModelSmall(null.model)
...
-    \item{model}{The model string.}
-    \item{outcome}{The outcome variable name.}
-    \item{covars}{A vector of covariate names}
 }
 
 The \code{fit} data frame contains the following columns, depending on which type of model was fit:
```
mconomos, line 118: @amstilp 
In lines 120-122, it looks like the descriptions got messed up.  

For `fitted.values`, it should say: 
"where X is the design matrix and beta is the vector of effect size estimates for the fixed effects"

Then for `resid.conditional` it should say:
"where `Z*u` represents the effects captured by the random effects specified with the cov.mat input."

man/fitNullModel.Rd
amstilp: Fixed

man/fitNullModel.Rd
amstilp: Thanks, fixed.

R/fitNullModel.R
amstilp: It looks like `.covMatNames`  doesn't return the matrix; it returns the sample.ids associated with the matrix. It also does some name checking to make sure that the rownames and colnames (for a given matrix in the list?) are the same. So I think it's ok for this update but we could look into whether any of the covMat code can be cleaned up in a new branch.

mconomos: sounds good

mconomos:
@amstilp sorry to be a pain, but question for you. How hard would it be to add `eta` (see e.g. line 137 of `runAIREMLgaussian`) to the `fit` object? Basically, `eta` is to `resid.conditional` what `fitted.values` is to `resid.marginal`. It seems like it might be more "complete" to include it as well. Sorry to just realize this now. It's not super important if it's a pain. 

If we can, I think we would call it `linear.predictor` in the `fit` object. 

The updates to the man page would be:

```
\item{fitted.values}{The fitted values from the model; i.e. X*beta where X is the design matrix and beta is the vector of effect size estimates for the fixed effects.}
\item{resid.marginal}{The marginal residuals from the model; i.e. Y - X*beta where Y is the vector of outcome values.}
\item{linear.predictor}{The linear predictor from the model; i.e. X*beta + Z*u, where Z*u represents the effects captured by the random effects specified with the cov.mat input.}
\item{resid.conditional}{The conditional residuals from the model; i.e. Y - X*beta - Z*u.}
```

R/prepareOutputForNullModel.R
```diff
@@ -182,11 +203,22 @@
     zeroFLAG <- vc.mod$zeroFLAG
     niter <- vc.mod$niter
 
-    out <- list(family = family, hetResid = hetResid, varComp = varComp,
+    # Sample-level data frame.
+    fit <- data.frame(
+      outcome = as.vector(y),
+      workingY = as.vector(workingY),
+      fitted.values = fitted.values,
+      resid.marginal = resid.marginal,
```
mconomos, line 208: @amstilp I think adding in `linear.predictor` might be as simple as including `vc.mod$eta` here

inst/NEWS.Rd
```diff
@@ -1,5 +1,15 @@
 \name{NEWS}
 \title{NEWS for GENESIS}
+\section{Version 2.19.9}{
```
smgogarten: This should actually be version 2.21.1 after last release

DESCRIPTION
```diff
@@ -3,7 +3,7 @@ Type: Package
 Title: GENetic EStimation and Inference in Structured samples
         (GENESIS): Statistical methods for analyzing genetic data from
         samples with population structure and/or relatedness
-Version: 2.19.8
+Version: 2.19.9
```
smgogarten: Should be version 2.21.1 (I just pulled the latest version number after release, sorry I forgot to do this earlier). Also update date.

man/fitNullModel.Rd
```diff
@@ -112,11 +110,29 @@ isNullModelSmall(null.model)
...
+  \item{resid.conditional}{The conditional residuals from the model; i.e. Y - X*beta - Z*u, where Z*u represents the effects captured by the random effects specified with the cov.mat input.}
+  \item{resid.cholesky}{The Cholesky residuals from the model; a transformation of the marginal residuals using the estimated model covariance structure such that they are uncorrelated and follow a standard multivariate Normal distribution with mean 0 and identity covariance matrix asymptotically. Linear regression of the Cholesky residual vector on an equivalently transformed genotype vector provides the same estimates as fitting the full GLS model (by the Frisch-Waugh-Lovell theorem).}
+  \item{resid.PY}{The outcome vector (Y) pre-multiplied by a projection matrix (P) that adjusts for covariates, random effects, and model covariance structure. These projected outcome values are essentially what are correlated with genotype values for association testing.}
+  \item{sample.id}{A vector of IDs for the samples used in the analysis, if called with an \code{AnnotatedDataFrame}.}
+}
```
smgogarten, line 129: `CX` and `CXCXI` are missing from the output description. @mconomos can you describe these elements? Should we just say "a matrix needed for calculating association" for both of them?

smgogarten (approved):
The `.R` scripts in the `vignettes` directory should not be in github. They are auto-generated from the `.Rmd` files.

DESCRIPTION
amstilp: Thanks, fixed!

inst/NEWS.Rd
amstilp: Fixed

amstilp:
I removed the R scripts in the vignettes directory, and updated the `.gitignore` file to prevent them from being added accidentally added in the future.

man/fitNullModel.Rd
mconomos: Good point. These have always been missing because there is no reason anyone would ever do anything with them. I think your suggested description is good.

R/prepareOutputForNullModel.R
amstilp: Done

man/fitNullModel.Rd
amstilp: I added these to the man page but changed the description to "a matrix needed for calculating association test statistics".

PR #59, opened 2020-12-16
Var approx v2

mconomos:
This definitely isn't complete, but I think I could use some feedback at this point. I'll annotate the code with some comments with specific questions I'd like input on.


R/assocTestSingle.R
```diff
@@ -93,10 +106,25 @@ setMethod("assocTestSingle",
 setMethod("assocTestSingle",
           "GenotypeIterator",
           function(gdsobj, null.model, test=c("Score", "Score.SPA", "BinomiRare", "CMP"),
-                   recalc.pval.thresh=0.05, GxE=NULL,
+                   recalc.pval.thresh=0.05, approx.score.SE=FALSE, GxE=NULL,
                    male.diploid=TRUE, verbose=TRUE) {
               test <- match.arg(test)
 
+              # Convert old null model format if necessary.
+              null.model <- .updateNullModelFormat(null.model)
```
mconomos, line 107: @amstilp I'm not sure why this wasn't here from your null model updates (like it is for SeqVarIterators). Was it just a mistake? or should I drop it?

R/testGeno.R
```diff
@@ -8,13 +8,13 @@
...
+                              recalc.pval.thresh = 1, approx.score.SE = FALSE, GxE.return.cov = FALSE){
     test <- match.arg(test)
     calc.score <- test %in% c("Score", "Score.SPA") | (recalc.pval.thresh < 1)
 
-    if (isNullModelSmall(nullmod) && (calc.score || !is.null(E))) {
-        stop("small null model cannot be used with options provided")
-    }
+    # if (isNullModelSmall(nullmod) && (calc.score || !is.null(E))) {
+    #     stop("small null model cannot be used with options provided")
+    # }
```
mconomos: I moved this check into assocTestSingle and integrated it with some other checks. If that's a problem, we can move them back. Also, this check had to be updated because you do (can) use a small null model when using approx.score.SE

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
+## estimate the ratio (score.se.ratio) of the true SE(score) to the fast SE(score)
+## update the null model to contain this parameter
+## this needs to be run before using the approx.score.se option in assocTestSingle
+## this is based off of the SAIGE variance approximation
+
```
mconomos: I don't love the name of this function because it's so long. Very open to other suggestions.

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+	
+     # calculate the score.se.ratio estimate
+     score.se.ratio <- .scoreSEratio(tab)
+
+     # update the null model
+     null.model <- nullModelSmall(null.model)
+     null.model$score.se.ratio <- score.se.ratio
+     return(null.model)
+}
+
```
mconomos: I tried to write this function in a way that it could be used as is for a user with a small dataset on e.g. a laptop, but with sub-functions that would be amenable to creating a workflow on SBG.

On SBG, I'm imagining that we would do this analysis across chromosomes by creating 22 jobs that would each run `.selectRandomVars` and then `.scoreSEtable`. We would then have another job that would rbind those 22 output tables and pass that into `.scoreSEratio` and update the null.model. 

I went in circles for a while on the best way to implement this. We don't _have to_ make an updated null model, we could instead just compute that `score.se.ratio` parameter and pass it into `assocTestSingle` separately. However, one of the advantages of this approximation is that you can use a small null model in the association test (though you need the full null model to run this part). Perhaps updating the null model should be a separate function? I'm undecided what is best. One disadvantage of the way I have it now is that the function doesn't return the table of estimates for each variant (output of `.scoreSEtable`). I think that table is useful for the keen user, but maybe something a less savvy user would just ignore. 

Anyways, I'm very open to discussing the best way in which to put these steps together to make them usable.

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+               out <- sample(out, size = nvar, replace = FALSE)
+               return(out)
+               })
+
+setMethod(".selectRandomVars",
+          "GenotypeData",
+          function(gdsobj, sample.id = NULL, nvar = 100, min.mac = 20, verbose = TRUE){
+
+               stop("function not yet implemented for GenotypeData objects")
+
```
mconomos: I started trying to put this together for a GenotypeData object, but it seems more challenging, and I'm not sure if it's worthwhile.

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+               return(tab)
+
+          })
+
+setMethod(".scoreSEtable",
+          "GenotypeData",
+          function(gdsobj, null.model, variant.id, male.diploid=TRUE, verbose = TRUE){
+
+               stop("function not implemented for GenotypeData objects")
+
```
mconomos: Same here. I think we just drop this commented out code if we decide it's unnecessary to implement for now.

R/assocTestSingle.R
amstilp: I think I just missed it. It's also missing in the `assocTestAggregate` method for `GenotypeIterator` objects. Can you add it there too?

R/testGeno.R
smgogarten: having it in assocTestSingle is better; we don't need to run the check for every iteration

R/assocTestSingle.R
```diff
@@ -15,6 +15,18 @@ setMethod("assocTestSingle",
...
+              if(calc.score && approx.score.SE && is.null(null.model$score.se.ratio)){
+              	stop("null.model must have score.se.ratio when approx.score.SE = TRUE")
+              } 
+              if(calc.score && !(approx.score.SE) && isNullModelSmall(null.model)){
+              	stop("small null.model cannot be used when approx.score.SE = FALSE")
+              }
+              if(!is.null(GxE) && isNullModelSmall(nullmod)){
+              	stop("small null.model cannot be used with GxE")
+              }
+
```
amstilp: You could turn this into an internal function (`.checkNullModelApproxScoreSE`?) -- it's easier to test.

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+               SeqArray::seqSetFilter(gdsobj, sample.id = sample.id, verbose = verbose)
+
+               out <- NULL
+               while(length(out) < nvar){
+                    # sample variants
+                    var.rand <- sort(sample.int(nvar.gdsobj, size = nvar, replace = FALSE))
+                    # filter to random sample
+                    SeqArray::seqSetFilter(gdsobj, variant.sel = var.rand, verbose = FALSE)
+                    # filter to MAC threshold
+                    if(min.mac > 0) SeqArray::seqSetFilterCond(gdsobj, mac = min.mac, verbose = FALSE)
```
smgogarten: have you checked that these filters are applied consecutively? (rather than `seqSetFilterCond` overriding the var.rand filter)

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+                    # collect selected variants
+                    out <- sort(unique(c(out, seqGetData(gdsobj, 'variant.id'))))
+               }
+
+               # sample down to number requested
```
smgogarten: didn't you already do this sampling above?

R/assocTestSingle.R
```diff
@@ -93,10 +106,25 @@ setMethod("assocTestSingle",
...
```
amstilp: See above comment about an internal function - since you run the checks in multiple places, it'll be easier to maintain the code longterm if you call an internal function for them.

R/assocTestSingle.R
```diff
@@ -93,10 +106,25 @@ setMethod("assocTestSingle",
 setMethod("assocTestSingle",
           "GenotypeIterator",
           function(gdsobj, null.model, test=c("Score", "Score.SPA", "BinomiRare", "CMP"),
-                   recalc.pval.thresh=0.05, GxE=NULL,
+                   recalc.pval.thresh=0.05, approx.score.SE=FALSE, GxE=NULL,
```
smgogarten: take out the `approx.score.SE=FALSE` argument from this method since it's not implemented. methods for different objects can have different sets of arguments.

R/updateNullModelApproxScoreSE.R
smgogarten: You don't even need to define a method with an error message, because if you try to call a generic function on an object that doesn't have a method implemented, you automatically get an error message like "no S4 method for this object type"

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+     # select a random set of variants meeting min.mac in the sample set
+     varid <- .selectRandomVars(gdsobj, sample.id = sampid, nvar = nvar, min.mac = min.mac, verbose = verbose)
+     if(verbose) message(paste('Selected', length(varid), 'variants with MAC >=', min.mac))
+
+     # calculate Score.SE using both approaches
+     tab <- .scoreSEtable(gdsobj, null.model, variant.id = varid, 
+                         sparse = sparse, imputed = imputed, male.diploid = male.diploid, genome.build = genome.build, verbose = verbose)
+	
+     # calculate the score.se.ratio estimate
+     score.se.ratio <- .scoreSEratio(tab)
```
mconomos: take a list of tables

mconomos: store the table in the null model object?

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
...
+     if(verbose) message(paste('null.model has', length(sampid), 'samples'))
+
+     # select a random set of variants meeting min.mac in the sample set
+     varid <- .selectRandomVars(gdsobj, sample.id = sampid, nvar = nvar, min.mac = min.mac, verbose = verbose)
+     if(verbose) message(paste('Selected', length(varid), 'variants with MAC >=', min.mac))
+
+     # calculate Score.SE using both approaches
+     tab <- .scoreSEtable(gdsobj, null.model, variant.id = varid, 
+                         sparse = sparse, imputed = imputed, male.diploid = male.diploid, genome.build = genome.build, verbose = verbose)
+
```
mconomos: split into two exported functions here.

R/updateNullModelApproxScoreSE.R
```diff
@@ -0,0 +1,246 @@
+## estimate the ratio (score.se.ratio) of the true SE(score) to the fast SE(score)
+## update the null model to contain this parameter
+## this needs to be run before using the approx.score.se option in assocTestSingle
+## this is based off of the SAIGE variance approximation
+
+updateNullModelApproxScoreSE <- function(gdsobj, null.model, nvar = 100, min.mac = 20, 
+     sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), verbose=TRUE){
+
```
mconomos: add optional variant.id input

R/assocTestSingle.R
mconomos: done.

R/assocTestSingle.R
mconomos: done

R/assocTestSingle.R
mconomos: added to assocTestAggregate

R/updateNullModelApproxScoreSE.R
mconomos: yes. they are applied consecutively

R/assocTestSingle.R
```diff
@@ -161,3 +173,23 @@ setMethod("assocTestSingle",
 
               as.data.frame(rbindlist(res))
           })
+
+
```
mconomos, line 178: created a null model check function for single variant tests

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
+## estimate the ratio (se.ratio) of the true SE(score) to the fast SE(score)
+## update the null model to contain this parameter
+## this needs to be run before using the approx.score.se option in assocTestSingle
+## this is based off of the SAIGE variance approximation
+
+
+setGeneric("fitNullModelApproxSE", function(x, ...) standardGeneric("fitNullModelApproxSE"))
+
```
mconomos: @smgogarten i'm not sure if all of these input arguments need to be specified in the method here or not. i always get a bit confused about that.

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
+## estimate the ratio (se.ratio) of the true SE(score) to the fast SE(score)
+## update the null model to contain this parameter
+## this needs to be run before using the approx.score.se option in assocTestSingle
+## this is based off of the SAIGE variance approximation
+
+
```
mconomos: this is the wrapper to: 
- fit the null model
- calculate the score and the approx at a set of variants
- update and return the small null model with the se.correction factor

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
...
+               null.model <- updateNullModApproxSE(null.model, tab)
+               return(null.model)
+          })
+
+
+calcScoreSE <- function(gdsobj, 
+                        null.model, 
+                        nvar = 100, 
+                        min.mac = 20, 
+                        variant.id = NULL,
```
mconomos: @amstilp added `variant.id` input for user specified variant list

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
...
+     message(paste('mean se.ratio: r = ', r))
+     message(paste('SE(r)/r = ', val))
+     if(val > 0.0025) warning(paste('It is recommended that SE(r)/r be < 0.0025. It is suggested to increase the number of variants in tab; try at least', 
+          round(nrow(tab)*(val/0.0025)^2, 0), 'variants'))
+
+     # update the null model
+     null.model <- nullModelSmall(null.model)
+
+     null.model$se.correction <- r
+     null.model$se.table <- tab
```
mconomos: added the table of SE calculations at each variant to the null model output

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
+## estimate the ratio (se.ratio) of the true SE(score) to the fast SE(score)
```
mconomos: I'm still stuck on good names. The to be exported functions are currently called:
- fitNullModelApproxSE
- calcScoreSE
- nullModelApproxSE

and the input parameter for `assocTestSingle` is called `approx.score.SE` (T/F). 

Another thought I had was to use "fast" instead of "approx". So maybe something like `fitNullModelFastScore`, `nullModelFastScore`, and `fast.score.SE`.

But I'm open to any name suggestions.

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,217 @@
...
+               # fit the null model
+               null.model <- fitNullModel(sampleData(x), ...)
+               # calculate score.SE and the fast approximation
+               tab <- calcScoreSE(gdsobj, null.model, ...)
+               # update the null model with the se.correction factor
+               null.model <- updateNullModApproxSE(null.model, tab)
+               return(null.model)
+          })
+
+
```
mconomos, line 42: Could also have this function return the `Score` value with minimal extra computation, and just call it `calcScore`. I wonder if this would have general utility...

R/nullModelApproxSE.R
```diff
@@ -0,0 +1,220 @@
...
+               out <- sample(out, size = nvar, replace = FALSE)
+               return(out)
+               })
+
+
+## function to calculate both the true and fast Score.SE for a specified set of variants
+## computes the ratio of the two Score.SE estimates for calculating the correction factor
+## this is called by updateNullModelApproxScoreSE
+setGeneric(".calcScore", function(gdsobj, ...) standardGeneric(".calcScore"))
+
```
mconomos: made it so that this function also returns the score value. might be useful if a user just wants to quickly compute the score for a few variants from a GDS file and null model.

R/fitNullModel.R
```diff
@@ -110,39 +125,6 @@ setMethod("fitNullModel",
           })
```
mconomos, line 127: moved this function into the nullModelInvNorm.R file with the other relevant functions

R/updateNullModOutcome.R
```diff
@@ -1,47 +1,50 @@
-
-### takes an existing linear model (family has to be "gaussian"!), rank-normalize residuals and scale, and re-fit null model. If group.idx are provided, optional scale by group.idx.
-
+## will update this file name to nullModelInvNorm.R
+## takes an existing null model (family has to be "gaussian"!), rank-normalize residuals and scale, and re-fit null model. 
+## If group.idx are provided, optional scale by group.idx.
 ## the names of items in the list group.idx have to match the names of the corresponding variance components!
```
mconomos: I moved the `updateNullModOutcome` code into `nullModelInvNorm` (which is exported). `nullModelInvNorm` did almost nothing outside of what `updateNullModOutcome` does, so I didn't see a good reason to keep them separate

R/nullModelInvNorm.R
```diff
@@ -0,0 +1,133 @@
+## will update this file name to nullModelInvNorm.R
```
mconomos, line 1: combined the nullModelInvNorm and updateNullModOutcome code here

R/updateNullModOutcome.R
```diff
@@ -1,124 +0,0 @@
-
```
mconomos, line 1: moved this code into nullModelInvNorm.R

tests/testthat/test_single_var_association_tests.R
```diff
@@ -201,9 +201,22 @@ test_that("small null model", {
...
-                                   recalc.pval.thresh=0.5),
-                 "small null model cannot be used")
-    expect_error(testGenoSingleVar(nullmod.small2, G = geno, test = "Score"),
-                 "small null model cannot be used")
+    # expect_error(testGenoSingleVar(nullmod.small2, G = geno, test = "BinomiRare",
+    #                                recalc.pval.thresh=0.5),
+    #              "small null.model cannot be used with test options provided")
+    # expect_error(testGenoSingleVar(nullmod.small2, G = geno, test = "Score"),
+    #              "small null.model cannot be used with test options provided")
+
```
mconomos, line 209: @amstilp can you please take a look at these. the checks below replaced the ones commented out above. we moved where those checks occur in the code. i'm not sure if these new checks should be a new "section" in the code, or just included where they are.

amstilp: I would change these to running `testGenoSingleVar` instead of `.checkNullModelTestSingle` to make sure that you're getting the correct error from `testGenoSingleVar` with the options provided, so for example test this:
```
    expect_error(testGenoSingleVar(nullmod.small2, G = geno, test = "BinomiRare",
        recalc.pval.thresh = 0.5, fast.score.SE = FALSE, GxE = NULL),
        "small null.model cannot be used with test options provided")
```

instead of this:

```
    expect_error(.checkNullModelTestSingle(nullmod.small2, test = "BinomiRare",
        recalc.pval.thresh = 0.5, fast.score.SE = FALSE, GxE = NULL),
        "small null.model cannot be used with test options provided")
```

That way it's harder to accidentally remove the step that checks the null model from the `testGenoSingleVar` function.

amstilp: Oh wait, I see. The checks happen in `assocTestSingle` instead. Ignore my above comment.

amstilp: In that case, I would make a new test specifically for testing `.checkNullModelTestSingle`. You could also add a check (or tests?) where it succeeds without an error. I think you can use expect_silent for that, or check that it returned an invisible NULL?

amstilp: Not a must have, since the regular single variant association tests aren't breaking. I'm a big fan of unit tests but I think it's fine to write them later given that check and the deadline. So let's leave this test as is for now and then write the more complete test later?

PR #63, opened 2021-03-03
Vsif

smgogarten:
@tamartsi the example and unit test are both failing because the `keep` vector in `sample.id=keep` is undefined. If I just take out that argument, the test fails because the sample.ids do not match. Do you just need to add a line of code defining `keep`?

Also, FYI, I merged the two man pages into one. I think it's easier for users to read about both functions on a single page, and there's less duplicated text.


smgogarten:
I have a question about name standardization: all other functions use `nullModel` in the name rather than `nullmod`. Should we change the name to `computeVSIFNullModel`, and the argument to `null.model`?

tamartsi:
@smgogarten agree about null model naming convention, also abut the man page. For some reason it worked on my computer without the keep, not sure why. 
Edit: now I know why, because I edited this and didn't clear my workspace (surprise surprise).
Should I updated my version with this and edit this or can you delete the sample.id = keep on your end?

tamartsi (changes requested):
in test_VSIF.R, to fix the error in the test of variant specific inflation factors using null model we should remove sample.id = keep (sorry about not checking thoroughly!).

R/computeVSIF.R
```diff
@@ -0,0 +1,117 @@
...
+  # define groups
+  groups <- colnames(eafs)
+  
+  # check that groups specificied by allele frequencies match
+  # the groups specified by group_var_vec
+  stopifnot(all(is.element(group_var_vec, groups)))
+  
+  
+  # subset group_var_vec (if needed) to match sample.ids in null.model 
+  if (!is.null(null.model$sample.id)){
```
mconomos, line 85: this needs to be updated to the new null model format. `null.model$fit$sample.id`

R/computeVSIF.R
```diff
@@ -0,0 +1,117 @@
...
+  ## prepare input for function computeVSIF
+  ns <- sigma_sqs <- rep(NA, length(groups))
+  names(ns) <- names(sigma_sqs) <- groups
+  
+  # extract marginal residuals from the null.model object:
+  if (!is.null(null.model$resid.marginal)){
+    resid <- null.model$resid.marginal
+    } else{
+    resid <- null.model$fit$resid.marginal
+  }
```
mconomos: rather than doing this check, i would just run `.updateNullModelFormat` at the beginning of the code

man/computeVSIF.Rd
```diff
@@ -0,0 +1,87 @@
...
+Computes variant-specific inflation factors resulting from differences 
+in variances and allele frequencies across groups pooled together in analysis.
+}
+\usage{
+computeVSIF(eafs, ns, sigma_sqs)
+computeVSIFNullModel(null.model, eafs, group_var_vec)
+}
+\arguments{
+  \item{eafs}{A named vector of a matrix of effect allele frequencies across
+  groups. Vector/column names are group names.}
```
mconomos: does it *have to* be effect allele frequencies? presumably it could be ref or alt allele freq just as well. is there another name that is more consistent with other GENESIS functions?

R/computeVSIF.R
```diff
@@ -0,0 +1,117 @@
...
+    colnames(eafs) <- names_eafs
+  }
+  
+  # define groups
+  groups <- colnames(eafs)
+  
+  # check that groups specificied by allele frequencies match
+  # the groups specified by group_var_vec
+  stopifnot(all(is.element(group_var_vec, groups)))
+
```
mconomos, line 84: run `.updateNullModelFormat` here

R/computeVSIF.R
```diff
@@ -0,0 +1,117 @@
...
+  # the groups specified by group_var_vec
+  stopifnot(all(is.element(group_var_vec, groups)))
+  
+  
+  # subset group_var_vec (if needed) to match sample.ids in null.model 
+  if (!is.null(null.model$sample.id)){
+    nullmod_sample.id <- null.model$sample.id
+  } else{ # there is not sample.id entry
+    nullmod_sample.id <- rownames(null.model$model.matrix)
+  }
```
mconomos, line 91: after running `.updateNullModelFormat`, I think this can just be `nullmod_sample.id <- null.model$fit$sample.id` (I think every null model has sample.id ...)

R/computeVSIF.R
smgogarten: I put it at the very beginning of the function instead

man/computeVSIF.Rd
smgogarten: renamed to `freq` for consistency with other functions

R/computeVSIF.R
smgogarten: not every null model has sample.id, but the ones that do have it under `fit`

PR #65, opened 2021-04-29
updated code for recessive coding to test in association tests

mchughc:
i've added code for assoc testing in a recessive setting, which basically just takes the genotypes (assumed to be coded additive) and transforms them. i haven't updated the analysis_pipeline code (need a new argument such as genotypeCoding or assocTestCoding or something) or done rigorous testing. but i hope it's a helpful start!!

### DIFF
--- a/R/assocTestSingle.R
+++ b/R/assocTestSingle.R
@@ -7,7 +7,8 @@ setMethod("assocTestSingle",
           function(gdsobj, null.model, test=c("Score", "Score.SPA", "BinomiRare", "CMP"),
                    recalc.pval.thresh=0.05, fast.score.SE=FALSE, GxE=NULL,
                    sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), 
-                   verbose=TRUE) {
+                   verbose=TRUE, geno.test=c('additive','dominant','recessive')) {
+
               test <- match.arg(test)
 
               # don't use sparse matrices for imputed dosages
@@ -18,7 +19,8 @@ setMethod("assocTestSingle",
 
               # check that the provided null model is compatible with the requested test
               .checkNullModelTestSingle(null.model = null.model, test = test, 
-              	recalc.pval.thresh = recalc.pval.thresh, fast.score.SE = fast.score.SE, GxE = GxE)
+                                        recalc.pval.thresh = recalc.pval.thresh, fast.score.SE = fast.score.SE, GxE = GxE)
+              ## could add a recessive + impute, etc. checking to this function
               
               # coerce null.model if necessary
               if (sparse) null.model <- .nullModelAsMatrix(null.model)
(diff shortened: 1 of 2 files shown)


R/assocTestSingle.R
```diff
@@ -7,7 +7,8 @@ setMethod("assocTestSingle",
           function(gdsobj, null.model, test=c("Score", "Score.SPA", "BinomiRare", "CMP"),
                    recalc.pval.thresh=0.05, fast.score.SE=FALSE, GxE=NULL,
                    sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"), 
-                   verbose=TRUE) {
+                   verbose=TRUE, geno.test=c('additive','dominant','recessive')) {
+
```
mconomos, line 11: I'm not sure that i want to use the term "test" again for the parameter here. How about `geno.coding`?

R/utils.R
```diff
@@ -60,18 +60,26 @@ setMethod("variantFilter",
...
-        mac <- round(pmin(count, 2*nsamp - count))
+        
+        if(genoTest=='additive'){
+            freq <- count/(2*nsamp)
+            mac <- round(pmin(count, 2*nsamp - count))
+        }else{ # recessive or dominant
+            freq <- count/nsamp
+            mac <- count
+        }
+
```
mconomos, line 82: For recessive, `freq` is the frequency of homozygous alts (i.e. the proportion of 1s) and `MAC` is the count of hom alts. 
For dominant, `freq` is the frequency of hets or hom alts (i.e. the proportions of 1s) and `MAC` is the count of hets and hom alts. 

@smgogarten and @amstilp - does this make sense to you? I think `freq` makes sense to me, but I'm less sure about `MAC`. Should `MAC` be the count of the smaller group? maybe that's more consistent with "minor". We're clearly abusing the name `MAC` here, but I don't think we want to deal with different names of the variable depending on the genotype coding, right?

R/utils.R
```diff
@@ -60,18 +60,26 @@ setMethod("variantFilter",
...
+        if(genoTest=='additive'){
+            freq <- count/(2*nsamp)
+            mac <- round(pmin(count, 2*nsamp - count))
+        }else{ # recessive or dominant
+            freq <- count/nsamp
+            mac <- count
+        }
+        
         return(data.frame(freq=freq, MAC=mac))
     }
```
mconomos, line 85: Noting that we need to address the `freq` and `MAC` calculations when `sex` is not null in the code below.

R/assocTestSingle.R
```diff
@@ -68,25 +79,27 @@ setMethod("assocTestSingle",
                       freq <- freq[keep,,drop=FALSE]
                   }
 
-                  # mean impute missing values
+                  ## mean impute missing values
                   if (any(n.obs < nrow(geno))) {
                       geno <- .meanImpute(geno, freq$freq)
```
mconomos, line 84: Need to adjust mean imputation when using recessive or dominant coding; use `freq` rather than `2*freq` in the `.meanImpute` function.

R/assocTestSingle.R
smgogarten: I like `geno.coding`

R/assocTestSingle.R
```diff
@@ -45,14 +47,23 @@ setMethod("assocTestSingle",
...
 
-                  # take note of number of non-missing samples
+                   # take note of number of non-missing samples
                   #n.obs <- colSums(!is.na(geno))
                   n.obs <- .countNonMissing(geno, MARGIN = 2)
 
+                  if(geno.test=='recessive'){ # is recessive
+                      ## if wanting to test a recessive model, the genotypes 0 and 1 are '0' and the genotype 2 is '1',
+                      ##eg. indicator of participant having 2 copies of the minor allele
+                      geno[geno==1] <- 0
```
smgogarten, line 57: Need to allow for missing values in the `geno` vector. I think using `%in%` instead of `==` will fix this.

R/utils.R
smgogarten: I don't think we should change the column name, and I think `MAC` should always count the smaller group. That would be `mac <- round(pmin(count, nsamp - count))`, right?

In the man page, we should include a section in `Details` listing things that are different for recessive and dominant coding, including how we defined `freq` and `MAC`.

R/assocTestSingle.R
amstilp: I also like `geno.coding`.

R/assocTestSingle.R
amstilp: Does it make more sense to impute the frequency before applying the recessive/dominant coding? Something like:
* f < 0.25: round to 0
* 0.25 < f < 0.75: code as 1 for recessive coding, code as 0 for dominant coding
* f > 0.75: code as 1 for recessive and dominant coding

Would probably need to think about the thresholds above, I'm not sure if they're appropriate. Does this even lead to different coding?

R/assocTestSingle.R
smgogarten: Turns out we should use `==` after all, as `%in%` doesn't work for `Matrix` objects.

smgogarten:
Implemented this feature on a different branch

PR #71, opened 2021-07-19
Parallel

smgogarten:
Loops in `pcrelate`, `assocTestSingle`, `assocTestAggregate`, and `admixMap` have been refactored to use the [BiocParallel](https://bioconductor.org/packages/release/bioc/html/BiocParallel.html) package. With BiocParallel, the user can define a parallel backend and either register it in the current R session or pass it as an argument to the GENESIS function, and that backend will be used to control parallelization. The default is Multicore, which uses one less than the available number of CPUs on the machine (of those, one is the head job so the actual number of workers is `N-2`).

We use the `bpiterate` function which takes two arguments: an iterator that supplies a new chunk of data every time it is called, and a function to run on that data (which can take additional arguments). `SeqVarIterator` objects are used to construct the iterators, which are defined separately within each method to return genotype data in the appropriate format. The functions called by `bpiterate` are shared across all methods for a given function (e.g., the `.testGenoBlockSingle` function is called by both `SeqVarIterator` and `GenotypeIterator` methods for `assocTestSingle`). This reduces the duplication of code compared to previous versions of GENESIS. `bpiterate` has the option of either returning a list containing all elements, or using a "reduce" function to combine elements. The association functions use list return and combine them with `rbindlist` as previously. `pcrelate` previously used a reduce function as implemented in the `foreach` package, and that has been replaced by the `bpiterate` call.

`bpiterate` returns errors as list elements, which is undesirable since all error messages then result in a failure of `rbindlist` to bind errors with successful output. This was addressed with the `.stopOnError` function that checks for errors and then prints them along with the data chunk in which they occurred.

Performance testing was done on a dataset with ~45k samples. I compared three scenarios: 1) previous GENESIS code, 2) BiocParallel with serial execution, 3) BiocParallel with multicore execution on a 24-core machine. In summary, the memory usage and run time is very similar between BiocParallel serial and the original code. BiocParallel multicore substantially reduces run time at the expense of additional memory.

assocTestSingle:
Scenario | Run time (s) | Memory (GB)
:--- | ---: | ---:
Original | 19369 | 5.6
BiocParallel serial | 20063 | 5.8
BiocParallel multicore | 4211 | 80

assocTestAggregate:
Scenario | Run time (s) | Memory (GB)
:--- | ---: | ---:
Original | 455 | 2.6
BiocParallel serial | 319 | 2.6
BiocParallel multicore | 266 | 38

pcrelate:
Scenario | Run time (s) | Memory (GB)
:--- | ---: | ---:
Original | 37049 | 28
BiocParallel serial | 36619 | 21
BiocParallel multicore | 13279 | 81


R/utils.R
```diff
@@ -58,14 +58,9 @@ setMethod("variantFilter",
...
 
-# index is in case we had to subset geno so it no longer matches the variant filter
-# (in the case of allele matching)
-.alleleFreq <- function(gdsobj, geno, variant.index=NULL, sample.index=NULL, male.diploid=TRUE, genome.build=c("hg19", "hg38")) {
-
-    # check sex
-    sex <- validateSex(gdsobj)
-    if (!is.null(sample.index)) sex <- sex[sample.index]
-    if (is.null(sex)) {
+.alleleFreq <- function(geno, chr=NULL, sex=NULL, male.diploid=TRUE) {
```
mconomos, line 103: it's nice that this function is simplified

R/nullModelFastScore.R
```diff
@@ -206,8 +206,9 @@ setMethod(".calcScore",
                n.obs <- .countNonMissing(geno, MARGIN = 2)
 
                # allele frequency
-               freq <- .alleleFreq(gdsobj, geno, sample.index=sample.index,
-                                        male.diploid=male.diploid, genome.build=genome.build)
+               chr <- chromWithPAR(gdsobj, genome.build=genome.build)
+               sex <- validateSex(gdsobj)[sample.index]
+               freq <- .alleleFreq(geno, chr, sex, male.diploid=male.diploid)
```
mconomos: @smgogarten - a lot of the code just above and below here is repeated from the new `.testGenoSingleBlock` function (count non-missing samples; allele freq; filter monomorphs; mean impute). is it worth the effort to make what's replicated its own sub function?

R/pcrelate.R
```diff
@@ -128,18 +125,30 @@ setMethod("pcrelate",
...
+            #G <- .readGeno(gdsobj, sample.include, snp.index = snp.blocks[[k]])
 
             # calculate ISAF betas
             beta <- .calcISAFBeta(G = G, VVtVi = VVtVi)
 
             # calculate PC-Relate estimates
             .pcrelateVarBlock(G = G, beta = beta, V = V, idx = 1:nrow(V), jdx = 1:nrow(V), scale = scale, ibd.probs = ibd.probs, maf.thresh = maf.thresh, maf.bound.method = maf.bound.method)
         }
+        
+        matList <- bpiterate(Giter, snpBlock, REDUCE = .matListCombine, reduce.in.order = FALSE, BPPARAM = BPPARAM)
```
mconomos, line 152: I think I probably just don't understand how `bpiterate` works, but why don't `VVtVi`, `V`, etc. need to be passed into the function here? I think the extra arguments were passed into `bpiterate` in the assoc functions.

man/admixMap.Rd
```diff
@@ -7,14 +7,16 @@ Run admixture analyses
...
+         genome.build=c("hg19", "hg38"), 
+         BPPARAM=bpparam(), verbose=TRUE)
 }
 
 \arguments{
   \item{admixDataList}{named list of \code{\link{GenotypeIterator}} or \code{\link{SeqVarIterator}} objects for each ancestry}
     \item{null.model}{A null model object returned by \code{\link{fitNullModel}}.}
   \item{male.diploid}{Logical for whether males on sex chromosomes are coded as diploid.}
   \item{genome.build}{A character sting indicating genome build; used to identify pseudoautosomal regions on the X and Y chromosomes.}
+  \item{BPPARAM}{A \code{\link{BiocParallelParam}} object to process blocks of variants in parallel.}
```
mconomos: should we say something about what the default does here? (and all the other pages)

mconomos:
@smgogarten - as far as I can tell, this all looks good. I left a couple minor comments/questions. 

Going back to one of my comments -- there's actually a fair amount of redundant code between `assocTestSingle` and `assocTestAggregate`, as well as `.testGenoBlockSingle` and `.testGenoBlockAggregate`. I'm not sure if it's worthwhile to try to turn those repeated bits into sub-functions themselves, or if that's a waste of time and effort.

R/nullModelFastScore.R
smgogarten: The same duplicated code is in assocTestAggregate. I made an internal function that can be used in all three places.

R/assocTestAggregate.R
```diff
@@ -251,11 +233,6 @@ setMethod("assocTestAggregate",
     res.var.i <- cbind(var.info, n.obs, freq, weight)
     
     if (n.site > 0) {
-        # mean impute missing values
```
mconomos: @smgogarten -- one difference i'm seeing with the new `.prepGenoBlock` function -- we're now doing the mean imputation _before_ computing `n.alt` and `n.sample.alt`. That is going to change those values. I'm not sure if it's a problem, but the value will be different.

R/assocTestAggregate.R
```diff
@@ -225,8 +209,6 @@ setMethod("assocTestAggregate",
         }
     } else {
         # user supplied weights
-        weight <- weight[keep]
-
```
mconomos: @smgogarten -- I think this `weight0` check could be build into the `.prepGenoBlock` function. Add a check on `if(!is.null(weight))` at line 60 of that function. Then we'd only have to do the subsets on `keep` once, instead of doing it again here. (double check my logic of the order of things)

R/assocTestAggregate.R
```diff
@@ -182,99 +133,116 @@ setMethod("assocTestAggregate",
...
+            weight <- weight[keep]
+        }
+    }
+    
+    # number of variant sites
+    n.site <- length(unique(var.info$variant.id))
+    
+    # number of alternate alleles
+    n.alt <- sum(geno, na.rm=TRUE)
+
```
mconomos, line 217: since we're imputing first, we have to change the way we get `n.alt`, but I think we can just use `sum(freq$MAC)` (actually should be faster anyways)

R/assocTestAggregate.R
```diff
@@ -182,99 +133,116 @@ setMethod("assocTestAggregate",
...
+    
+    # number of variant sites
+    n.site <- length(unique(var.info$variant.id))
+    
+    # number of alternate alleles
+    n.alt <- sum(geno, na.rm=TRUE)
+    
+    # number of samples with observed alternate alleles > 0
+    n.sample.alt <- sum(rowSums(geno, na.rm=TRUE) > 0)
+
```
mconomos, line 220: since we're imputing first, we have to change the way we get `n.sample.alt`. this one is a bit trickier, but here's what I came up with:
`n.sample.alt <- sum(.apply(geno, 1, function(x){ any(x >= 1) }))`

Basically, check by row (sample) if any values are at least 1. I'm assuming that any imputed value would be < 1. This assumption will hold as long as the MAF != 0.5, which would be unlikely for rare variant tests.

R/assocTestAggregate.R
smgogarten: Ah, that's why we were doing it later in the aggregate test. We could just move the `meanImpute` step outside of `.prepGenoBlock`, that would be more straightforward than computing those quantities in different ways.

R/pcrelate.R
smgogarten: In the assoc functions, the function called by `bpiterate` is defined outside of the `assocTest*` function, so it can be used by multiple methods. Here, it's defined inside the `pcrelate` function, so all those variables are already in scope.

R/assocTestAggregate.R
smgogarten: `n.alt` is not necessarily the same as `sum(freq$MAC)`, because the alternate allele could be major.

R/assocTestAggregate.R
mconomos: That's also fine with me.

R/assocTestAggregate.R
mconomos: Ah. Good point. I always forget that we code the alt allele and not the minor allele in the rare variant tests (I think all other software does the minor allele). 

The "cumulative MAC" is actually a more useful measure than `n.alt`... should we consider changing it? They're _usually_ the same, but not always, as you pointed out.

R/assocTestAggregate.R
smgogarten: Good idea, I've done that

R/assocTestAggregate.R
smgogarten: MAC is reported in the `variantInfo` for each aggregate unit, so a user could get cumulative MAC pretty easily by summing over that.

man/admixMap.Rd
smgogarten: I'm hesitant to give too much detail in case BiocParallel changes how the `bpparam` function behaves, but I included a link to the `bpparam` documentation. (Notably, it does something different depending on the OS, as Windows doesn't support Multicore.)

R/assocTestAggregate.R
mconomos: You are correct. I still think cumulative MAC would be useful in the main results table -- e.g. I think analysts often only share that table with collaborators, not all the unit specific tables. But for the sake of not changing things, it's fine to leave it as is.

mconomos:
@smgogarten - as far as I can tell, this all seems good to go!

PR #75, opened 2021-08-20
recessive and dominant coding

smgogarten:
Add the option for recessive or dominant coding in `assocTestSingle`. Additional columns will be returned with number of samples who are coded "1" in each case (`n.hom.alt` and `n.any.alt` respectively).


R/utils.R
```diff
@@ -167,10 +189,14 @@ setMethod("variantFilter",
...
 
-.meanImputeFn <- function(geno, freq) {
+.meanImputeFn <- function(geno, freq, geno.coding="additive") {
     miss.idx <- which(is.na(geno))
     miss.var.idx <- ceiling(miss.idx/nrow(geno))
-    geno[miss.idx] <- 2*freq[miss.var.idx]
+    if (geno.coding == "additive") {
+        geno[miss.idx] <- 2*freq[miss.var.idx]
+    } else {
+        geno[miss.idx] <- freq[miss.var.idx]
```
mconomos, line 209: @smgogarten I actually don't think this is correct. I think we want to impute with `freq[[out.col]]/n.obs` -- i.e. the frequency of the "1" value. Unfortunately, that seems like it might mean greater modification to this function.

man/assocTestSingle.Rd
```diff
@@ -31,6 +34,7 @@
...
     \item{GxE}{A vector of character strings specifying the names of the variables for which a genotype interaction term should be included.If \code{GxE} is not \code{NULL}, \code{test} is ignored and Wald tests of interaction are performed. If \code{GxE} is \code{NULL} (default) no genotype interactions are included. See 'Details' for further information.}
     %\item{ivar.return.betaCov}{Logical indicator of whether the estimated covariance matrix of the effect size estimates (betas) for the genotype and genotype interaction terms should be returned; the default is FALSE.}
+    \item{geno.coding}{Whether genotypes should be coded as "additive" (0, 1, or 2 copies of the effect allele), "recessive" (1=homozygous for the effect allele, 0 otherwise), or "dominant" (1=heterozygous or homozygous for the effect allele, 0 for no effect allele).}
```
mconomos: @smgogarten - should we say "effect" or "alternate" here? We say alternate for the `freq` output.

smgogarten: For `GenotypeData` objects, it's alternate only if the "A" allele is alternate. Maybe we should update the `freq` description.

mconomos: I always forget this! Yes, we should probably update the `freq` description then. 

We actually use "alternate allele" in a lot of places in this documentation where we mean "alternate allele if it's a SeqVarData object" and "A allele if it's a GenotypeData object". We should probably clarify that somehow. In line 51 of the Rd file, we make this distinction. Maybe we should just define "effect allele" there (to be the alt or A allele in each case), and then use the term effect allele throughout the rest of the document?

R/utils.R
smgogarten: Can we just get that frequency from `geno` directly as `freq <- colMeans(geno, na.rm=TRUE)`? We don't do that in the additive case because 1) we've just calculated frequency already and 2) sex chromosomes need special treatment, but I don't think either of those apply here.

mconomos: Yeah. That should be fine - just compute the `freq` vector in the else statement right here. 

Do sex chr need any special consideration for recessive or dominant coding? Females should be fine. For males, a carrier should definitely be a "1" for dominant (I think they would be); a carrier should probably(?) be a "1" for recessive, which they might not be? (I'm actually not sure).

smgogarten: Males are generally coded as 0/2, in which case a carrier would be "1" for either dominant or recessive. We do have the option for a user to specify `male.diploid=FALSE`, in which case they are expected to be coded 0/1. So I think we need another `if` statement under recessive, where if `male.diploid=FALSE` then X chr variants for males get set to 1 if `geno == 1`.

smgogarten: My thinking was the same as Adrienne's. For a recessive trait, carriers are not affected (or are less affected) because they still have a copy of the non-effect allele being expressed. For males on the X chromosome, if they have the effect allele they will exhibit the trait. (Hemophilia for example, which is caused by a recessive X chromosome mutation.) The additive/recessive/dominant coding reflects the biology of the trait, while the `male.diploid` flag only reflects the conventions used in the input VCF file.

mconomos: Works for me. We should just make sure the documentation is clear.

man/assocTestSingle.Rd
```diff
@@ -71,6 +75,10 @@
     \item{n.obs}{The number of samples with non-missing genotypes}
     \item{freq}{The estimated alternate allele frequency}
     \item{MAC}{The minor allele count. For multiallelic variants, "minor" is determined by comparing the count of the alternate allele specified by \code{allele.index} with the sum of all other alleles.}
+    If \code{geno.coding} is \code{"recessive"}:
+    \item{n.hom.alt}{The number of samples homozygous for the effect (alternate) allele.}
+    If \code{geno.coding} is \code{"dominant"}:
+    \item{n.any.alt}{The number of samples with any copies of the effect (alternate) allele.}
```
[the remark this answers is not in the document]
smgogarten: Should we then call this `n.hom.eff` and `n.any.eff`? Similarly, in `assocTestAggregate` we have `n.alt` and `n.sample.alt`, but I wouldn't expect those to be used as often with array data. (Also, changing established output names now might break code and cause more confusion than it's worth).

man/assocTestSingle.Rd
```diff
@@ -31,6 +34,7 @@
...
     %\item{ivar.return.betaCov}{Logical indicator of whether the estimated covariance matrix of the effect size estimates (betas) for the genotype and genotype interaction terms should be returned; the default is FALSE.}
+    \item{geno.coding}{Whether genotypes should be coded as "additive" (0, 1, or 2 copies of the effect allele), "recessive" (1=homozygous for the effect allele, 0 otherwise), or "dominant" (1=heterozygous or homozygous for the effect allele, 0 for no effect allele).}
     \item{sparse}{Logical indicator of whether to read genotypes as sparse Matrix objects; the default is \code{TRUE}. Set this to \code{FALSE} if the alternate allele dosage of the genotypes in the test are not expected to be mostly 0.}
     \item{imputed}{Logical indicator of whether to read dosages from the "DS" field containing imputed dosages instead of counting the number of alternate alleles.}
     \item{male.diploid}{Logical for whether males on sex chromosomes are coded as diploid.}
```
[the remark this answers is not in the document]
smgogarten: I think that doesn't apply based on above discussion, but I added clarification on what this argument means.

R/utils.R
smgogarten: There's even a word for it! Males are hemizygous on the X chromosome.

man/assocTestAggregate.Rd
```diff
@@ -60,7 +60,7 @@
   association test, even if a different set of samples is present in the
   current filter for \code{gdsobj}.
   
-    The effect size estimate is for each copy of the alternate allele (when \code{gdsobj} is a \code{\link{SeqVarIterator}} object) or the "A" allele (when \code{gdsobj} is a \code{\link{GenotypeIterator}} object). For multiallelic variants, each alternate (or "A") allele is tested separately.
+    The effect size estimate is for each copy of the alternate allele (when \code{gdsobj} is a \code{\link{SeqVarIterator}} object) or the "A" allele (when \code{gdsobj} is a \code{\link{GenotypeIterator}} object). For multiallelic variants in \code{\link{SeqVarIterator}} objects, each alternate (or "A") allele is tested separately.
```
mconomos: @smgogarten after the sentence:

"The effect size estimate is for each copy of the alternate allele (when \code{gdsobj} is a \code{\link{SeqVarIterator}} object) or the "A" allele (when \code{gdsobj} is a \code{\link{GenotypeIterator}} object)."

Should we add something like:

"We refer to this as the "effect allele" in the rest of the documentation."

mconomos:
I left one minor comment on the documentation, otherwise, this seems like it's ready to go?

PR #76, opened 2021-10-01
functions for meta analysis

mconomos:


R/meta.R
```diff
@@ -0,0 +1,247 @@
...
+
+setMethod("metaPrepScores",
+          "SeqVarIterator",
+          function(gdsobj, null.model, score.cov = TRUE,
+                   geno.coding=c("additive", "dominant", "recessive"),
+                   sparse=TRUE, imputed=FALSE, male.diploid=TRUE, genome.build=c("hg19", "hg38"),
+                   BPPARAM=bpparam(), verbose=TRUE) {
+
+                     # better way to handle these checks?
+                     if(class(gdsobj) == 'SeqVarBlockIterator' && score.cov){
```
smgogarten: should be `if(is(gdsobj, "SeqVarBlockIterator") && score.cov)`

R/utils.R
```diff
@@ -23,10 +23,15 @@ setMethod("validateSex",
 setMethod("variantInfo",
           "GenotypeData",
           function(gdsobj, alleles=FALSE) {
-              data.frame(variant.id=getSnpID(gdsobj),
+              x <- data.frame(variant.id=getSnpID(gdsobj),
                          chr=getChromosome(gdsobj, char=TRUE),
                          pos=getPosition(gdsobj),
                          stringsAsFactors=FALSE)
+              if (alleles) {
```
smgogarten, line 30: I just made a change in GWASTools so this should work now (allele columns were missing previously)

PR #78, opened 2021-10-11
Subset covariance matrix to samples in sample.id if specified

amstilp:
* When fitting a null model with an AnnotatedDataFrame, `fitNullModel` will subset `cov.mat` to the samples specified in `sample.id` before checking for consistency with `x`.
* Add a unit test for this behavior

Is this too confusing of a change to make?

_Edited after some additional commits_: this pull request also modifies the null model code such that all samples in the `x` AnnotatedDataFrame are required to be in `cov.mat`. If `sample.id` is passed, subset both `x` and `cov.mat` to the samples in `sample.id`.

### DIFF
--- a/R/fitNullModel.R
+++ b/R/fitNullModel.R
@@ -18,9 +18,9 @@ setMethod("fitNullModel",
                    drop.zeros = TRUE,
                    return.small = FALSE,
                    verbose = TRUE) {
-              
+
               family <- .checkFamily(family, two.stage)
-              
+
               if (is.data.table(x)) x <- as.data.frame(x)
 
               desmat <- createDesignMatrix(x, outcome, covars, group.var)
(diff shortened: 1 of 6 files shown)


R/fitNullModel.R
```diff
@@ -89,6 +85,10 @@ setMethod("fitNullModel",
                   }
               }
 
+              if (!is.null(cov.mat)) {
+                  .checkSampleId(cov.mat, x)
+              }
+
```
amstilp: The only real change in this function is to move this check after the subsetting by `sample.id`.

smgogarten: Will this still work if `cov.mat` does not have rownames and colnames specified correctly (which `.checkSampleId` previously ensured)?

smgogarten (changes requested):
If we don't have one already, we need a unit test for the case when `sample.id` is provided AND `cov.mat` is missing row and column names, as I suspect that could fail in the new code. We might need to split up the two checks in `checkSampleId` so one happens before the `sample.id` code block and the other happens after.

R/fitNullModel.R
amstilp: It does still work, but the reason isn't intuitive from just looking at the code. I think the covariance matrix gets subset to a zero by zero matrix when attempting to subset by sample id. Splitting the check into two makes more sense.

amstilp:
I'm still wondering if it makes more sense to require that all samples in `x` are in the covariance matrix (unless a `sample.id` argument is passed). It would require people calling `fitNullModel` with a `SeqVarData` (or iterator) object to explicitly set `sample.id`, but that's less confusing to me than specifying the set of samples by the rownames/colnames of the covariance matrix.

amstilp:
I ended up making the change to require `cov.mat` to have all samples in `x` unless `sample.id` is passed. We can choose not to keep it if it's not the desired behavior.

smgogarten:
I agree that it makes more sense this way. I'm going to merge this and get it in Bioconductor for the next release (deadline is tomorrow).

PR #81, opened 2022-01-06
replace travis build with github actions

smgogarten:


.github/workflows/check-bioc.yml
```diff
@@ -0,0 +1,291 @@
...
+## * OS: operating system
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
```
amstilp: You may want to add `- main` here also, in case we ever switch the `master` branch to `main`. It's not necessary to do it now (and it could be done if the branch name changes), but it won't hurt anything and it would be easy to forget in the future.

.github/workflows/check-bioc.yml
```diff
@@ -0,0 +1,291 @@
...
+    runs-on: ${{ matrix.config.os }}
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+    container: ${{ matrix.config.cont }}
+    ## Environment variables unique to this job.
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - { os: ubuntu-latest, r: 'devel', bioc: '3.15', cont: "bioconductor/bioconductor_docker:devel", rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest" }
```
amstilp, line 61: Does `bioc: '3.15'` indicate the current devel version of bioconductor? Because the bioconductor version numbers change, it might be better to specify it differently (if you can somehow specify "release" or "devel" instead of a version number), or add it as an environment variable at the top (under `run docker`).

It looks like there is a [different but related version](https://github.com/lcolladotor/biocthis/blob/e6e5cb3e122c3112784fcd04536df9ed50021a30/actions/check-bioc.yml) of `check-bioc.yaml` that might do some of this (the `findrversion` step), but I'm not sure if you started there and had to change some of it for some reason.

amstilp: Oh hm, the different version I linked to appears to be from an old commit in `biocthis` and I haven't tracked down why all the changes happened. It still seems odd to point to a specific version -- won't it have to be updated every time there's a new Bioconductor release?

smgogarten: I wondered this also, but I used the current version of `biocthis::use_bioc_github_action(biocdocker='devel')` and this is what it gave me. I guess it's just something we'll have to remember to update twice a year.

smgogarten: I wondered this also, but I used the current version of `biocthis::use_bioc_github_action(biocdocker='devel')` and this is what it gave me. I guess it's just something we'll have to remember to update twice a year.

.github/workflows/check-bioc.yml
smgogarten: Good point. Renaming `master` to `main` is on my to-do list, but I know Matt is in the middle of working on multiple feature branches and I don't want to create additional headaches for his local environment right now.

PR #85, opened 2022-02-17
Fix bug with SPA p-values for glm models

### LINKED ISSUE #84: SPA association tests are incorrect for non-mixed models
`fitted.values` in for glm null models already have `expit` applied, but we are treating them as if they do not in GENESIS. This leads to incorrect (and inflated) association test statistics.

amstilp:
In `SPA_pval`, the fitted.values for glm models already have the expit applied, so taking the expit again causes incorrect association test statistics.

### DIFF
--- a/man/fitNullModel.Rd
+++ b/man/fitNullModel.Rd
@@ -34,7 +34,7 @@
 
   \code{isNullModelSmall} returns TRUE if a null model is small; FALSE otherwise.
 
-  \code{fitNullModelFastScore} fits a null model that can be used for association testing with the fast approximation to the score standard error (SE). 
+  \code{fitNullModelFastScore} fits a null model that can be used for association testing with the fast approximation to the score standard error (SE).
 
   \code{calcScore} calculates the score, its true SE, and the fast SE for a set of variants; used to compute the SE correction factor used for the fast approximation.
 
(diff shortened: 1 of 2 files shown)


man/fitNullModel.Rd
```diff
@@ -34,7 +34,7 @@
 
   \code{isNullModelSmall} returns TRUE if a null model is small; FALSE otherwise.
 
-  \code{fitNullModelFastScore} fits a null model that can be used for association testing with the fast approximation to the score standard error (SE). 
+  \code{fitNullModelFastScore} fits a null model that can be used for association testing with the fast approximation to the score standard error (SE).
```
amstilp, line 37: @mconomos is this a reasonable update to the docs?

amstilp:
I am still testing this, but pushed the branch and started a pull review for commenting.

amstilp:
After running `Score.SPA` tests using one of my glm models with the new code, it looks much better - assoc test statistics have a much better QQ plot.

man/fitNullModel.Rd
mconomos: @amstilp - I'd say it slightly differently to address other families with other link functions (e.g. poisson regression).

\item{fitted.values}{The fitted values from the model. For mixed models, this is X*beta where X is the design matrix and beta is the vector of effect size estimates for the fixed effects. For non-mixed models, this is the inverse link function applied to X*beta (e.g. expit(X*beta) for logistic regression when \code{family = "binomial"}.}

amstilp: Thanks! Updated.

PR #90, opened 2022-04-18
add fastSMMAT option and document

mconomos:


mconomos:
@smgogarten I actually think this might be all that needs to be changed to "release" fastSMMAT. Please read my description to see if it makes sense to you.

smgogarten (approved):
Looks good

PR #93, opened 2022-06-02
Add burden test values to SMMAT output

amstilp:
Add burden ~~and SKAT~~ test statistics to the SMMAT output.

### DIFF
--- a/R/variantSetTests.R
+++ b/R/variantSetTests.R
@@ -136,7 +136,8 @@ testVariantSet <- function( nullmod, G, weights,
     G.rowSums <- rowSums(G) # P^{1/2}GW1
     GG1 <- crossprod(G, G.rowSums) # WGPGW1  # O(mn)
     V.sum <- sum(GG1) # 1WGPGW1
-    burden.pval <- pchisq(U.sum^2/V.sum, df=1, lower.tail=FALSE)
+    burden.stat = U.sum^2 / V.sum
+    burden.pval <- pchisq(burden.stat, df=1, lower.tail=FALSE)
 
     # adjust U and G for burden
     U <- U - GG1*U.sum/V.sum # WGPY - WGPGW1 * 1WGPY/(1WGPGW1)
@@ -168,6 +169,9 @@ testVariantSet <- function( nullmod, G, weights,
         smmat.pval <- burden.pval
     }
     return(list(
+      Score_burden = U.sum,
+      Score.SE_burden = sqrt(V.sum),
+      Stat_burden = burden.stat,
       pval_burden = burden.pval,
       pval_theta = theta.pval,
       pval_SMMAT = smmat.pval,
--- a/R/variantSetTests.R
+++ b/R/variantSetTests.R
@@ -136,8 +136,8 @@ testVariantSet <- function( nullmod, G, weights,
     G.rowSums <- rowSums(G) # P^{1/2}GW1
     GG1 <- crossprod(G, G.rowSums) # WGPGW1  # O(mn)
     V.sum <- sum(GG1) # 1WGPGW1
-    burden.stat = U.sum^2 / V.sum
-    burden.pval <- pchisq(burden.stat, df=1, lower.tail=FALSE)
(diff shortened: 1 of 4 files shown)


amstilp:
@mconomos To add SKAT output, I think we'd have to run either `.testVariantSetSKAT` before adjusting `U` and `G` for the burden. Is that correct? If so, is it worth the extra computation?

mconomos:
@amstilp No, I don't think it's worth adding in the SKAT output

R/variantSetTests.R
```diff
@@ -167,7 +168,20 @@ testVariantSet <- function( nullmod, G, weights,
         err <- 1
         smmat.pval <- burden.pval
     }
-    return(list(pval_burden = burden.pval, pval_theta = theta.pval, pval_SMMAT = smmat.pval, err = err, pval_theta.method = out$pval.method))
+    return(list(
+      # burden test output.
+      Score_burden = U.sum,
+      Score.SE_burden = sqrt(V.sum),
+      Stat_burden = burden.stat,
```
mconomos, line 175: @amstilp -- for consistency with the regular burden test output, the Stat should probably be the sqrt of what you have here. i.e. U.sum/sqrt(V.sum). I also can't really decide if we even need to include the stat here. The burden test does return Score.Stat, but it's so easy to get from the Score and the SE...

man/assocTestAggregate.Rd
```diff
@@ -114,7 +114,11 @@
     \item{pval.method}{The p-value calculation method used. When standard SKAT is used, one of "integration" or "saddlepoint"; when fastSKAT random projections are used to approximate eigenvalues of the genotype covariance matrix, one of "ssvd_integration" or "ssvd_saddlepoint"; when fastSKAT random projections are used to approximate both the eigenvalues and the trace of the genotype covariance matrix, one of "rsvd_integration" or "rsvd_saddlepoint".}
 
     If \code{test} is \code{"SMMAT"} or \code{"fastSMMAT"}:
+    \item{Score_burden}{The value of the score function for the burden test}
+    \item{Score.SE_burden}{The estimated standard error of the Score for the burden test}
+    \item{Stat_burden}{The score Z test statistic for the burden test}
     \item{pval_burden}{The burden test p-value}
+    \item{Q_theta}{The SMMAT test statistic}
```
mconomos: @amstilp Q_theta: The test statistic for the adjusted SKAT test (which is asymptotically independent of the burden test)

tests/testthat/test_variant_set.R
```diff
@@ -44,25 +44,44 @@ test_that("SMMAT matches burden", {
 	nullmod <- .testNullmod(n, MM=TRUE)
         smmat <- .testVariantSetSMMAT(nullmod, geno, weights)
         burden <- .testVariantSetBurden(nullmod, geno, weights, burden.test="Score")
+        expect_equal(smmat$Score_burden,burden$Score)
+        expect_equal(smmat$Score.SE_burden, burden$Score.SE)
+        # Burden test does not return a test statistic.
```
mconomos: @amstilp -- I think it does. It's `Score.Stat`. But it's the sqrt of what you had implemented in the SMMAT version

amstilp: Wow, how did I miss that 🤦‍♀️

man/assocTestAggregate.Rd
amstilp: Updated

R/variantSetTests.R
amstilp: Fixed by taking the sqrt of burden.stat. Might as well return it?

mconomos: Actually, we should probably compute burden.stat as U.sum/sqrt(V.sum) and then square when calculating the p-value. The way it is now, we lose the sign of the test statistic. Sorry, didn't think of this til right now.

amstilp: oh! good point. fixed.

mconomos (approved):
I think this all looks good to go. Thanks!

PR #94, opened 2022-06-07
enable returning the covariance matrix for GxE tests

mconomos:


mconomos:
@smgogarten -- I enabled returning the GxE covariance matrix to help Ingo out in this branch. Pretty much all the code was already there, just commented out. If you have 10 min to look at it, that'd be great. I'm fine with leaving this as a separate branch for now and pointing him here to get the code if that's preferable.

smgogarten (commented):
It's not that simple: if `GxE.return.cov` is set to TRUE, the value returned by `testGenoSingleVar` is a list instead of a data.frame, and everything downstream will break.
https://github.com/UW-GAC/GENESIS/blob/master/R/testGeno.R#L199

We would need at minimum if statements [here](https://github.com/UW-GAC/GENESIS/blob/master/R/assocTestSingle.R#L174), [here](https://github.com/UW-GAC/GENESIS/blob/master/R/assocTestSingle.R#L69) and [here](https://github.com/UW-GAC/GENESIS/blob/master/R/assocTestSingle.R#L131), and the whole structure of the object returned by `assocTestSingle` would have to change: no longer a data.frame but a list containing a data.frame and a list of matrices. We would also need to write unit tests for this case.

PR #111, opened 2024-03-19
Add a helper function to calculate and filter pchisq values

### LINKED ISSUE #110: Filter extreme p-values using `.Machine$double.xmin`
In https://github.com/UW-GAC/analysis_pipeline assoc_plots, instead of hardcoding the threshold for p-value to 38.5, use `.Machine$double.xmin` to calculate the threshold that can be used. The current code is:

```
if ("stat" %in% names(assoc)) {
    extreme <- abs(assoc$stat) > 38.5
    if (any(extreme)) assoc$pval[extreme] <- 5e-324
}
```

5e-324 was probably chosen based on this information from ?.Machine:
```
     Note that on most platforms smaller positive values than
     ‘.Machine$double.xmin’ can occur.  On a typical R platform the
     smallest positive double is about ‘5e-324’.
```

but it's most reliable to use the value in .Machine. The code would then change to:
```
if ("stat" %in% names(assoc)) {
    extreme <- assoc$stat^2 > qchisq(.Machine$double.xmin, df=1, lower.tail=F)
    if (any(extreme)) assoc$pval[extreme] <- .Machine$double.xmin
}
```

A p-value of 5e-324 and eg 2.225074e-308 (from .Machine$double.xmin on my machine) are basically the same, anyway.

amstilp:
Filter p-values based on the Machine$double.xmin value: if a p-value calculated with pchisq is smaller than Machine$double.xmin, then set it to the double.xmin value.

This should prevent pval=0 situations.

Closes #110

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -352,3 +352,10 @@ setMethod(".annotateAssoc",
     put.attr.gdsn(geno.node, "sample.order")
     closefn.gds(gds)
 }
+
+.pchisq_filter_extreme <- function(...) {
+    # Note: stat is the squared test statistic.
+    pval = pchisq(...)
+    pval[pval < .Machine$double.xmin] = .Machine$double.xmin
+    return(pval)
+}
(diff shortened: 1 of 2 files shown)


amstilp:
Closing - branched off of wrong branch

PR #112, opened 2024-03-19
Filter extremely small p-values by setting them to Machine$double.xmin

amstilp:
For any p-values calculated by pchisq, set any p-values smaller than Machine$double.xmin to the double.xmin value. This should help prevent pval=0 cases.

Closes #111

### DIFF
--- a/R/testSingleVarianceComponent.R
+++ b/R/testSingleVarianceComponent.R
@@ -25,7 +25,7 @@ testSingleVarianceComponent <- function(nullmod, varCompName, covMatList, group.
 
 	test.stat <- -2*(nullmod.noVarComp$logLik - nullmod$logLik)
 
-	pval <- 0.5*pchisq(test.stat, df = length(nullmod$varComp), lower.tail = FALSE) + 0.5*pchisq(test.stat, df = length(nullmod$varComp) - 1, lower.tail = F)
+	pval <- 0.5*.pchisq_filter_extreme(test.stat, df = length(nullmod$varComp), lower.tail = FALSE) + 0.5*.pchisq_filter_extreme(test.stat, df = length(nullmod$varComp) - 1, lower.tail = F)
 
 	return(pval)
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -354,8 +354,10 @@ setMethod(".annotateAssoc",
 }
 
 .pchisq_filter_extreme <- function(...) {
-    # Note: stat is the squared test statistic.
+    args <- list(...)
     pval = pchisq(...)
-    pval[pval < .Machine$double.xmin] = .Machine$double.xmin
+    if ("df" %in% names(args) && args$df > 0) {
+        pval[pval < .Machine$double.xmin] = .Machine$double.xmin
+    }
     return(pval)
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -352,3 +352,10 @@ setMethod(".annotateAssoc",
     put.attr.gdsn(geno.node, "sample.order")
     closefn.gds(gds)
 }
+
+.pchisq_filter_extreme <- function(...) {
+    # Note: stat is the squared test statistic.
+    pval = pchisq(...)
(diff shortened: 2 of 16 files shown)


R/testSingleVarianceComponent.R
```diff
@@ -25,7 +25,7 @@ testSingleVarianceComponent <- function(nullmod, varCompName, covMatList, group.
 
 	test.stat <- -2*(nullmod.noVarComp$logLik - nullmod$logLik)
 
-	pval <- 0.5*pchisq(test.stat, df = length(nullmod$varComp), lower.tail = FALSE) + 0.5*pchisq(test.stat, df = length(nullmod$varComp) - 1, lower.tail = F)
+	pval <- 0.5*.pchisq_filter_extreme(test.stat, df = length(nullmod$varComp), lower.tail = FALSE) + 0.5*.pchisq_filter_extreme(test.stat, df = length(nullmod$varComp) - 1, lower.tail = F)
```
mconomos, line 28: @amstilp -- I believe, for this test, it's possible for the df=0 for the reduced model, in which case the actual p-value for that component is 0. I think this might be a problem for your .pchisq_filter_extreme function

amstilp: Are the p-values returned by pchisq always 0 if df=0?

mconomos: Unless the test-statistic == 0, then the p-value = 1. 
A chi-square df = 0 distribution is a point mass at 0

amstilp: Ok. I can change it to something like this:
```
.pchisq_filter_extreme <- function(...) {
    args <- list(...)
    # Note: stat is the squared test statistic.
    pval = pchisq(...)
    if ("df" %in% names(args) && args$df > 0) {
        pval[pval < .Machine$double.xmin] = .Machine$double.xmin
    }
    return(pval)
}
```

which should cover the df=0 case?

mconomos: @amstilp -- I think "df" is a required argument, so do you really need to check if it's in the list?

amstilp: Good point - I rearranged the code a bit and the first check is no longer necessary. Will fix.

inst/NEWS.Rd
```diff
@@ -2,6 +2,12 @@
 
 \title{NEWS for GENESIS}
 
+\section{Version 2.33.2}{
+  \itemize{
+    \item Set extremely small p-values calculated with pchisq to \code{Machine$double.xmin}.
```
mconomos: @amstilp I think we could be more clear here that the issue is being returned as 0, not that we just picked some threshold for some reason. Maybe: 

Set extremely small p-values returned as 0 by pchisq to \code{Machine$double.xmin}.

amstilp: I would put it like this: "Set extremely small p-values (\code{< Machine$double.xmin}) calculated with pchisq to \code{Machine$double.xmin}. This change prevents GENESIS from returning p-values equal to 0."

because it's not just pval=0; it's any p-values less than Machine$double.xmin. So you could have pchisq return 1e-324 but then the .pchisq_filter_extreme would change it to the Machine$double.xmin value.

mconomos: Makes sense and works for me. Everything else about this update looks good to me.

R/utils.R
```diff
@@ -352,3 +352,12 @@ setMethod(".annotateAssoc",
     put.attr.gdsn(geno.node, "sample.order")
     closefn.gds(gds)
 }
+
+.pchisq_filter_extreme <- function(...) {
+    args <- list(...)
+    pval = pchisq(...)
```
smgogarten, line 358: just curious... is there a reason for using `=` instead of `<-` here?

smgogarten (approved):
Looks great, thanks

R/utils.R
amstilp: No reason, probably just an effect of going between R and python!

